PR #7, opened 2022-02-06
refactor: Add refactored helper modules

adigitoleo:
Add modules for handling:
- deformation mechanisms
- crystalline fabrics
- experimental stiffnesses
- tensor operations

### DIFF
--- a/src/pydrex/deformation_mechanism.py
+++ b/src/pydrex/deformation_mechanism.py
@@ -1,11 +1,4 @@
-"""PyDRex: Deformation mechanism enums.
-
-Public symbols:
-- `Regime`
-
-"""
-__all__ = ["Regime"]
-
+"""PyDRex: Deformation mechanism enums."""
 from enum import Enum, unique
 
 
--- a/src/pydrex/deformation_mechanism.py
+++ b/src/pydrex/deformation_mechanism.py
@@ -1,0 +1,17 @@
+"""PyDRex: Deformation mechanism enums.
+
+Public symbols:
+- `Regime`
+
+"""
+__all__ = ["Regime"]
+
+import enum as e
+
+
+@e.unique
+class Regime(e.Enum):
+    diffusion = 0
+    dislocation = 1
+    byerlee = 2
+    max_viscosity = 3
--- a/src/pydrex/deformation_mechanism.py
+++ b/src/pydrex/deformation_mechanism.py
@@ -6,11 +6,11 @@ Public symbols:
 """
 __all__ = ["Regime"]
 
-import enum as e
+from enum import Enum, unique
 
 
-@e.unique
-class Regime(e.Enum):
+@unique
+class Regime(Enum):
     diffusion = 0
     dislocation = 1
     byerlee = 2
--- a/src/pydrex/stiffness.py
+++ b/src/pydrex/stiffness.py
@@ -1,0 +1,39 @@
+"""PyDRex: Stiffness matrices for minerals.
+
+Currently, stiffness matrices are provided for olivine and enstatite:
+- `OLIVINE`
+- `ENSTATITE`
+
+The source of the values used here is unknown, but they are copied
+from the original DRex code:
+
+<http://www.ipgp.fr/~kaminski/web_doudoud/DRex.tar.gz> [88K download]
+
+The provided matrices encode the stiffness tensor in the compact Voigt notation.
+
+"""
+__all__ = ["OLIVINE", "ENSTATITE"]
+import numpy as np
+
+
+OLIVINE = np.array(
+    [
+        [320.71, 69.84, 71.22, 0.0, 0.0, 0.0],
+        [69.84, 197.25, 74.8, 0.0, 0.0, 0.0],
+        [71.22, 74.8, 234.32, 0.0, 0.0, 0.0],
+        [0.0, 0.0, 0.0, 63.77, 0.0, 0.0],
+        [0.0, 0.0, 0.0, 0.0, 77.67, 0.0],
+        [0.0, 0.0, 0.0, 0.0, 0.0, 78.36],
+    ]
+)
+
+ENSTATITE = np.array(
+    [
+        [236.9, 79.6, 63.2, 0.0, 0.0, 0.0],
+        [79.6, 180.5, 56.8, 0.0, 0.0, 0.0],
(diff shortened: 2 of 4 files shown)


adigitoleo:
First step of merging in my changes. This is mostly just preperatory stuff. I'll put things into src/pydrex/ because that will allow the old code to coexist at the moment, and also it's easier to turn into a package that way.

`deformation_mechanisms` is new, `fabrics` used to be in params, same as the stiffness matrices. The tensor ops used to be in the main code, I'm moving them out for clarity and later maybe they can be replaced/merged with e.g. pytensors or those other packages we looked up a while ago.

adigitoleo:
"Those other packages":
- https://github.com/romerogroup/MechElastic
- https://github.com/MahdiDavari/FPTE
- https://github.com/libAtoms/matscipy
- https://github.com/materialsproject/pymatgen

The last two are more for atomic-level (polycrystal) modelling, so probably less relevant. Pytensors seems to have disappeared, the old [Pypi page is still up](https://pypi.org/project/pytensors/) but the repo link is dead.

adigitoleo:
In my hunt for where the stiffness values come from, I followed a trail of references back to Ahrens 1995 "Mineral Physics & Crystallography: A Handbook of Physical Constants" (AGU), here is the table for olivine from in there:

image

Values seem to be kind of similar for the 300K case, which is a bit worrying because we probably want to be using 1500K values...

adigitoleo:
Enstatite values are even more of a mystery. Browaeys & Chevrot 2004 also have some values for both olivine and enstatite in section 4.1, and they are different again.

src/pydrex/stiffness.py
```diff
@@ -0,0 +1,39 @@
...
+Currently, stiffness matrices are provided for olivine and enstatite:
+- `OLIVINE`
+- `ENSTATITE`
+
+The source of the values used here is unknown, but they are copied
+from the original DRex code:
+
+<http://www.ipgp.fr/~kaminski/web_doudoud/DRex.tar.gz> [88K download]
+
+The provided matrices encode the stiffness tensor in the compact Voigt notation.
```
Patol75, line 12: Mention entries are elastic constants in GPa?

src/pydrex/tensors.py
```diff
@@ -0,0 +1,106 @@
...
+                    delta_rs = 1 if r == s else 0
+                    j = (r + 1) * delta_rs + (1 - delta_rs) * (7 - r - s) - 1
+                    matrix[i, j] += tensor[p, q, r, s]
+                    matrix_indices[i, j] += 1
+    matrix /= matrix_indices
+    return (matrix + matrix.transpose()) / 2
+
+
+def voigt_matrix_to_vector(matrix):
+    """Create the 21-component Voigt vector equivalent to the 6x6 Voigt matrix."""
```
Patol75, line 75: Let's be consistent with the spelling of Voigt. Capital V probably makes sense.

src/pydrex/tensors.py
```diff
@@ -0,0 +1,106 @@
...
+    return vector
+
+
+@nb.jit(nopython=True)
+def rotate(tensor, rotation):
+    rotated_tensor = np.zeros((3, 3, 3, 3))
+    for i in range(3):
+        for j in range(3):
+            for k in range(3):
+                for l in range(3):
```
Patol75, line 94: Using `l` as a variable is ambiguous. I let you judge.

Patol75 (changes requested):
Breaking down the code into distinct separate files makes sense, thank you for doing it. I left a few minor items for you to look at.

src/pydrex/deformation_mechanism.py
```diff
@@ -0,0 +1,17 @@
+"""PyDRex: Deformation mechanism enums.
+
+Public symbols:
+- `Regime`
+
+"""
+__all__ = ["Regime"]
```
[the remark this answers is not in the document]
adigitoleo: This is not strictly necessary, the only effects are:
- `from foo import *` only imports names from `__all__` [1]
- some documentation and linting tools can rely on this

For example pylint gives an error if something from `__all__` is missing.
I was going to trial this as a way to distinguish public vs internal interface.
Most python standard library modules use this.
However, it does require typing things twice which is a bit annoying.
Let me know if we should rather remove it.

This could be made use of in `__init__.py` files when creating a package, but there are other options.

src/pydrex/deformation_mechanism.py
```diff
@@ -0,0 +1,17 @@
...
+
+import enum as e
```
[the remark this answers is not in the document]
adigitoleo: Yes I'll change it. Comes from a habit of being paranoid about import conflicts, and https://gist.github.com/sloria/7001839#imports.

src/pydrex/tensors.py
adigitoleo: You mean with capital `I`? OK I can change it back, wasn't sure.

src/pydrex/deformation_mechanism.py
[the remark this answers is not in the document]
Patol75: I do not think the code benefits from it, and considering it is annoying to do in the first place, probably best to remove it if you agree; `from foo import *` is not good practice anyway. We will see if it is useful when we deal with `__init__.py` files.

src/pydrex/tensors.py
Patol75: https://www.flake8rules.com/rules/E741.html

src/pydrex/deformation_mechanism.py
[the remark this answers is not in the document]
adigitoleo: OK, I'll also get rid of the `Public symbols` noise, I think I included it for documentation but `help(module)` already prints a list of functions:

```
FUNCTIONS
    Voigt_matrix_to_vector(matrix)
        Create the 21-component Voigt vector equivalent to the 6x6 Voigt matrix.

    Voigt_to_elastic_tensor(matrix)
        Create 4-th order elastic tensor from an equivalent Voigt matrix.

    elastic_tensor_to_Voigt(tensor)
        Create a 6x6 Voigt matrix from an equivalent 4-th order elastic tensor.

    rotate(tensor, rotation)
        Rotate 4-th order tensor using a 3x3 rotation matrix.

    upper_tri_to_symmetric(arr)
        Create symmetric array using upper triangle of input array.

```

Patol75 (approved):

PR #8, opened 2022-02-08
refactor: Add input handlers

adigitoleo:
Adds modules to handle VTK and parameter input.
More tests should be added in time, but for now some simple mock files are added in data/.
To see how the parameter parsing works, run `parse_params("data/pydrex.ini")`.
Configuration file format is INI, because that's what Python's configparser understands.

### DIFF
--- a/src/pydrex/io.py
+++ b/src/pydrex/io.py
@@ -1,0 +1,141 @@
+"""PyDRex: Input/Output functions and helpers."""
+import configparser
+import pathlib
+import runpy
+
+import numpy as np
+
+import pydrex.exceptions as _err
+
+
+def parse_params(file):
+    """Parse an INI file containing PyDRex parameters."""
+    config = configparser.ConfigParser()
+    configpath = _resolve_path(file)
+    config.read(configpath)
+
+    geometry = config["Geometry"]
+    output = config["Output"]
+    params = config["Parameters"]
+
+    mesh = read_mesh(_resolve_path(geometry.get("meshfile"), configpath.parent))
+    olivine_fraction = params.getfloat("olivine_fraction")
+    enstatite_fraction = params.getfloat("enstatite_fraction", 1.0 - olivine_fraction)
+
+    # TODO: Allow user-given order for Euler angle output (like Fraters 2021)?
+    olivine_output = tuple(
+        output.get("olivine", fallback="volume_fraction, rotation_matrix")
+        .replace(" ", "")
+        .split(",")
+    )
+    enstatite_output = tuple(
+        output.get("enstatite", fallback="volume_fraction, rotation_matrix")
+        .replace(" ", "")
+        .split(",")
+    )
+
+    return dict(
+        zip(
+            [
+                "mesh",
+                "simulation_name",
+                "checkpoint_interval",
+                "olivine_output",
+                "enstatite_output",
+                "olivine_fraction",
+                "enstatite_fraction",
+                "stress_exponent",
+                "gbm_mobility",
+                "gbs_threshold",
+                "nucleation_efficiency",
+                "olivine_fabric",
+                "number_of_grains",
+            ],
+            [
+                mesh,
+                output.get("simulation_name", fallback="simulation_name"),
+                output.getint(
+                    "checkpoint_interval", fallback=0
+                ),  # 0 disables checkpointing
+                ["olivine_" + s for s in olivine_output],  # Unique names for np.savez()
+                [
+                    "enstatite_" + s for s in enstatite_output
+                ],  # Unique names for np.savez()
+                olivine_fraction,
+                enstatite_fraction,
+                params.getfloat("stress_exponent"),
+                params.getfloat("gbm_mobility"),
+                params.getfloat("gbs_threshold"),
+                params.getfloat("nucleation_efficiency"),
+                params.get("olivine_fabric"),
+                params.getint("number_of_grains"),
+            ],
+        )
+    )
+
+
+def _resolve_path(path, refdir=None):
+    cwd = pathlib.Path.cwd()
+    if refdir is None:
+        _path = cwd / path
+    else:
+        _path = cwd / refdir / path
+    if _path.is_file():
+        return _path.resolve()
+    raise IOError(f"file '{_path}' does not exist")
+
+
+def read_mesh(meshfile):
+    """Read mesh from `meshfile` into numpy arrays.
+
+    Supported formats:
+    - Python script that assigns grid coordinates (2D Numpy array, see `create_mesh`)
+
+    """
+    if meshfile.suffix == ".py":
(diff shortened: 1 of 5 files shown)


adigitoleo:
Custom exceptions might be overkill, I added them before finding meshio (they would be useful if we needed to roll our own mesh parsers, but meshio might do the job here).

src/pydrex/io.py
```diff
@@ -0,0 +1,141 @@
...
+            [
+                mesh,
+                output.get("simulation_name", fallback="simulation_name"),
+                output.getint(
+                    "checkpoint_interval", fallback=0
+                ),  # 0 disables checkpointing
+                ["olivine_" + s for s in olivine_output],  # Unique names for np.savez()
+                [
+                    "enstatite_" + s for s in enstatite_output
+                ],  # Unique names for np.savez()
```
Patol75, line 63: This list should be formatted similarly to the previous one.

src/pydrex/io.py
```diff
@@ -0,0 +1,141 @@
...
+
+def read_mesh(meshfile):
+    """Read mesh from `meshfile` into numpy arrays.
+
+    Supported formats:
+    - Python script that assigns grid coordinates (2D Numpy array, see `create_mesh`)
+
+    """
+    if meshfile.suffix == ".py":
+        return create_mesh(runpy.run_path(meshfile)["gridcoords"])
```
Patol75, line 96: I am wondering if `importlib` should be used instead of `runpy`, but I am probably worrying for no reason.

src/pydrex/io.py
```diff
@@ -0,0 +1,141 @@
...
+                "gridcoords",
+                "gridmin",
+                "gridmax",
+            ],
+            [
+                dimension,
+                [arr.size for arr in gridcoords],
+                gridcoords,
+                [np.amin(arr) for arr in gridcoords],
+                [np.amax(arr) for arr in gridcoords],
```
Patol75, line 138: `arr.min()` and `arr.max()` should be sufficient?

src/pydrex/vtk_helpers.py
```diff
@@ -0,0 +1,57 @@
...
+    Throws a LookupError if `fieldname` does not match any data.
+
+    """
+    data = points.GetAbstractArray(fieldname)  # Silently returns None if missing...
+    if data is None:
+        raise LookupError(f"unable to find data for {fieldname}")
+
+    return np.array([data.GetTuple(i) for i in range(data.GetNumberOfTuples())])[
+        :, :dimensions
+    ]
```
Patol75, line 40: Does this work for tensor fields?

Patol75 (changes requested):
New files look sensible. A few comments for you to look at.

src/pydrex/io.py
adigitoleo: Ah, that's due to running [black](https://pypi.org/project/black/). I'll change it, don't need to repeat the comment anyway.

src/pydrex/io.py
adigitoleo: I think for importlib it has to be a proper module, with an `__init__.py` or something:

```
>>> import importlib as i
>>> i.import_module("data/2d_mesh_test.py")
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    i.import_module("data/2d_mesh_test.py")
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'data/2d_mesh_test'
```

adigitoleo: I agree that having the "mesh file" be executable is not ideal, it would be best to read from an established mesh format like GMSH (or maybe something more stable...)

Patol75: Something like that works, but it is far from a one-liner! Haha
```
import importlib
import sys

spec = importlib.util.spec_from_file_location('my_script_name',
                                              'path_to_my_script')
my_script = importlib.util.module_from_spec(spec)
sys.modules['my_script_name'] = my_script
spec.loader.exec_module(my_script)

print(my_script.my_variable)
```

src/pydrex/vtk_helpers.py
adigitoleo: Yes but actually no, I was using it to read e.g. "VelocityGradient" by passing `dimensions = d ** 2` where `d` is 2 or 3. However, this doesn't actually work as I expected (in 2D), because the third column is always full of zeros. I'll need to fix it, do you know if that is specific to fluidity? Or does VTK always use a third dimension full of zeros.

adigitoleo: OK so I read something about how the third dimension is needed to make things play nice with paraview. That's kinda annoying, since a column of zeros is ambiguous (is it actual data? just a dummy array to appease paraview?) I had this problem when refactoring the original code for my shearbox models because I have a unidirectional velocity field, so something like https://github.com/Patol75/PyDRex/blob/master/src/PyDRex.py#L720 would not work. I will change the code to assume that the third dimension is zero for 2D models.

Patol75: Why would it not work?

adigitoleo: I mean that I can't use that to detect the dummy columns (not in the coords array but in the others), but anyway I think I've got something cleaner incoming.

src/pydrex/io.py
adigitoleo: I'm sticking with `runpy` here!

src/pydrex/vtk_helpers.py
```diff
@@ -24,34 +24,84 @@ def get_output(filename):
...
     data = points.GetAbstractArray(fieldname)  # Silently returns None if missing...
     if data is None:
         raise LookupError(f"unable to find data for {fieldname}")
 
-    return np.array([data.GetTuple(i) for i in range(data.GetNumberOfTuples())])[
-        :, :dimensions
-    ]
+    a = np.array([data.GetTuple(i) for i in range(data.GetNumberOfTuples())])
+    if skip_z:
+        return _skip_column_every(a, 3)
```
adigitoleo, line 80: Should work for tensors, from what I understand they are stored in VTK as tuples with n > dims components, so e.g. VelocityGradient is a list of tuples with 9 components each in 3D models.

Patol75 (approved):

PR #9, opened 2022-02-09
refactor: Add interpolation callbacks and helpers.

adigitoleo:
Adds some functions to organise interpolator callbacks.
This was previously handled in `main`. At the moment the option to use
matplotlib interpolation is handled in command line args.
It could make sense to move that into the configuration file later.

### DIFF
--- a/src/pydrex/interpolations.py
+++ b/src/pydrex/interpolations.py
@@ -1,9 +1,12 @@
 """PyDRex: Interpolation callbacks and helpers."""
 # NOTE: Module contains delayed imports (search for 'Delayed import' comments).
-import importlib
 import itertools as it
 
 import numpy as np
+
+from scipy.interpolate import CloughTocher2DInterpolator
+from matplotlib.tri import CubicTriInterpolator
+from scipy.interpolate import NearestNDInterpolator
 
 import pydrex.exceptions as _err
 import pydrex.vtk_helpers as _vtk
@@ -46,6 +49,7 @@ def default_interpolators(config, coords, vtk_output, mpl_interp=None):
 
     velocity = _vtk.read_tuple_array(data, "Velocity", skip_z=is_2d)
     velocity_gradient = _vtk.read_tuple_array(data, "VelocityGradient", skip_z=is_2d)
+
     if is_2d:
         fields = (
             "velocity_x",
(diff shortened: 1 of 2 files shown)


adigitoleo:
Initially I put the callbacks in a NamedTuple, but that didn't play nice with Charm, because it uses pickle to send objects between Python processes and pickle doesn't know how to sereialize NamedTuples... Anyway, dicts is fine so long as we don't touch the callbacks after creation.

adigitoleo:
Bleh, missing some imports and stuff, hang on.

src/pydrex/interpolations.py
```diff
@@ -0,0 +1,281 @@
...
+
+    Supported interpolator choices (dim = `coords.shape[1]`):
+
+    |                                 |   scalar data   |   vector data   |
+    |       interpolator              |    and dim ==   |    and dim ==   |
+    |                                 |   2    |   3    |   2    |   3    |
+    |---------------------------------------------------------------------|
+    | "CloughTocher2DInterpolator" [1]|        |        |   x    |        |
+    | "CubicTriInterpolator" [2]      |        |        |   x    |        |
+    | "NearestNDInterpolator" [3]     |   x    |   x    |        |   x    |
```
Patol75, line 171: Why do you distinguish between scalar and vector? Those interpolators only work for scalar data, or am I missing something?

Patol75 (changes requested):
Looks fine to me, just one comment for you to look at.

src/pydrex/interpolations.py
adigitoleo: Ah, it's not very clear. By "scalar" I meant a vector of scalars, see https://github.com/Patol75/PyDRex/pull/9/files/b29d96fac5159cacfb8bb0c9e0980af5349bc8dd#diff-4281486f0b06da1f6a2e3a09a74550f908b703735f2974f31632131bc0f81806R207

What I'm trying to do is always use NearestNDInterpolator for things like the DeformationMechanism field, equivalent to https://github.com/Patol75/PyDRex/blob/master/src/PyDRex.py#L773

adigitoleo: Maybe there is a better interpolator class to use here.

adigitoleo: Looks like deformation mechanism always has shape `(n,1)` where `n` is the number of nodes:

```
In [5]: defmech = _vtk.read_tuple_array(pointdata, "DeformationMechanism")

In [6]: defmech
Out[6]:
array([[1.],
       [1.],
       [1.],
       ...,
       [1.],
       [1.],
       [1.]])

In [7]: defmech[0]
Out[7]: array([1.])

In [8]: defmech.shape
Out[8]: (1584, 1)
```

Maybe it's best to modify the `is_scalar` condition to not take arrays of shape `(n,)`.

Patol75: Wait, this does not have anything to do with shape. The deformation mechanism is a discrete field, so you need to do an interpolation based on the nearest neighbour algorithm, otherwise, you end up with meaningless values. For continuous fields, I use the Clough-Tocher scheme in 2-D because it is fast enough. In 3-D, I fall back to the nearest neighour because the linear interpolation scheme is usually too slow.

adigitoleo: OK I guess I was assuming that the only scalar field input would be deformation mechanism. I'll remove the confusing validator and change the docs.

Patol75 (approved):

PR #12, opened 2022-02-19
fix: Fix unnecessary use of list comprehension

adigitoleo:


src/pydrex/vtk_helpers.py
```diff
@@ -45,7 +45,7 @@ def _skip_column_every(a, n):
     array([], shape=(10, 0), dtype=float64)
     
     """
-    return a[:, np.mod([i for i in range(a.shape[1])], n) != n - 1]
+    return a[:, np.mod(list(range(a.shape[1])), n) != n - 1]
```
Patol75, line 48: `np.arange(a.shape[1])`

Patol75 (changes requested):

src/pydrex/vtk_helpers.py
adigitoleo: Thanks, fixed.

Patol75 (approved):

PR #14, opened 2022-02-20
refactor: Rename diagnostic output options

adigitoleo:


adigitoleo:
So that it matches the names in `core` better.

Patol75 (approved):

PR #18, opened 2022-02-21
fix: Fix strain energy calculation for enstatite

adigitoleo:
Don't raise a possibly negative number to a power

### DIFF
--- a/src/pydrex/core.py
+++ b/src/pydrex/core.py
@@ -523,9 +523,7 @@ def _get_slip_rate_softest(deformation_rate, velocity_gradient):
             enumerator += 2 * deformation_rate[j, L] * velocity_gradient[j, L]
             denominator += 2 * deformation_rate[j, L] ** 2
 
-        slip_rate_softest = enumerator / denominator
-
-    return slip_rate_softest
+    return enumerator / denominator
 
 
 @nb.njit(fastmath=True)
@@ -628,10 +626,7 @@ def _get_strain_energy(
                 -nucleation_efficiency * dislocation_density ** 2
             )
     elif np.all(rrss == _fabric.RRSS_ENSTATITE):
-        # TODO: What is this? Why is it required?
-        weight_factor = slip_rate_softest * rrss[slip_indices[0]] ** (-stress_exponent)
-        # TODO: Verify rrss[i] == τ_0 / τ^sv
-        assert rrss[slip_indices[-1]] == 1
+        weight_factor = slip_rate_softest / rrss[slip_indices[-1]] ** stress_exponent
         dislocation_density = rrss[slip_indices[-1]] ** (
             1.5 - stress_exponent
         ) * np.abs(weight_factor) ** (1.5 / stress_exponent)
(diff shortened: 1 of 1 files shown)


adigitoleo:
slight change from original fortran because I don't understand why line 1437 is necessary if tau_ens is always 1

src/pydrex/core.py
```diff
@@ -628,16 +626,16 @@ def _get_strain_energy(
                 -nucleation_efficiency * dislocation_density ** 2
             )
     elif np.all(rrss == _fabric.RRSS_ENSTATITE):
-        # TODO: What is this? Why is it required?
-        weight_factor = slip_rate_softest * rrss[slip_indices[0]] ** (-stress_exponent)
```
Patol75, line 632: Wait, isn't it just `rrss[slip_indices[-1]]`?

src/pydrex/core.py
```diff
@@ -628,16 +626,16 @@ def _get_strain_energy(
...
         # TODO: Verify rrss[i] == τ_0 / τ^sv
         assert rrss[slip_indices[-1]] == 1
-        dislocation_density = rrss[slip_indices[-1]] ** (
+        dislocation_density = 1 ** (
             1.5 - stress_exponent
-        ) * np.abs(weight_factor) ** (1.5 / stress_exponent)
+        ) * np.abs(slip_rate_softest) ** (1.5 / stress_exponent)
```
Patol75, line 633: Potentially not needed based on the above.

Patol75 (changes requested):

src/pydrex/core.py
adigitoleo: Yeah, that was the first mistake but then I realised that I haven't seen anyone use a value other than 1 for the rrss on the only active slip system of enstatite. However, I suppose it makes more sense to use `slip_indices` if we're already doing that for olivine.

adigitoleo: Based on the fortran I think it is `1 / rrss[slip_indices[-1]]` so that's what I've done in the ammended commit

Patol75: Careful, it is either `(1 / rrss[slip_indices[-1]]) ** stress_exponent` or `rrss[slip_indices[-1] ** -stress_exponent`

Patol75 (changes requested):

src/pydrex/core.py
```diff
@@ -628,16 +626,15 @@ def _get_strain_energy(
                 -nucleation_efficiency * dislocation_density ** 2
             )
     elif np.all(rrss == _fabric.RRSS_ENSTATITE):
-        # TODO: What is this? Why is it required?
-        weight_factor = slip_rate_softest * rrss[slip_indices[0]] ** (-stress_exponent)
-        # TODO: Verify rrss[i] == τ_0 / τ^sv
-        assert rrss[slip_indices[-1]] == 1
+        weight_factor = slip_rate_softest * (1 / rrss[slip_indices[-1]]) ** (-stress_exponent)
```
Patol75, line 629: `weight_factor = slip_rate_softest / rrss[slip_indices[-1]] ** stress_exponent`

Patol75 (changes requested):

Patol75 (approved):

PR #21, opened 2022-02-21
fix: Stop integration if velocity is too small

adigitoleo:


src/pydrex/core.py
```diff
@@ -91,6 +91,10 @@ def solve(config, interpolators, node):
         logging.debug("max. velocity gradient: %s", velocity_gradient.max())
         logging.debug("min. velocity gradient: %s", velocity_gradient.min())
 
+        if la.norm(velocity, ord=2) < 1e-15:
+            logging.info("stopping integration prematurely, velocity magnitude is too small")
+            break
+
```
Patol75, line 97: I think you could have `velocity_norm = la.norm(velocity, ord=2)` within or after where you call the interpolants and use that variable in lines 90, 94 and 109 to avoid calculating the same thing 3 times.

Patol75 (changes requested):

Patol75 (approved):

PR #23, opened 2022-02-22
feat: Re-enable shared-memory multiprocessing

adigitoleo:
First step of bringing multiprocessing back to the refactored code.


adigitoleo:
`PIDFileHandler` from `run.py` (for setting up log files for each process) is not working with this, that will be something to debug later.

Patol75 (approved):

PR #27, opened 2022-03-09
refactor: Overhaul core routines (again), add Mineral class

adigitoleo:
- Pathline stuff moved to its own file
- Introducing `Mineral`, to represent a mineral aggregate
- RK4 scheme is now a member of `Mineral` (`update_orientations`)
- Begginings of a test suite, still very work-in-progress

Reasons for change:
- The pathline way of doing things gets fiddly for a number of cases, e.g. zero velocity
- The pathline way is only useful for steady-state flow
- Refactoring onto a `Mineral` class makes it easier to run for single particles,
  e.g. a particle tracer
- Refactoring onto a `Mineral` class means that some nice metadata can be stored there
- The `core.derivatives` can now be jitted (maybe small speedups)

What is a `Mineral`? This idea is inspired by Fraters 2021.
A particle can store multiple minerals, which in turn store multiple grains.
So `Mineral` basically provides a container for grains, i.e. an aggregate.
In DRex, mineral phases don't interact and they rotate independently.


adigitoleo:
Also `Enum` -> `IntEnum` to appease numba.

src/pydrex/core.py
```diff
@@ -1,49 +1,103 @@
...
+    Update crystalline orientations and grain volume distribution
+    for minerals undergoing plastic deformation.
+    Only 3D simulations are currently supported.
 
-def solve(config, interpolators, node):
+    Args:
+    - `minerals` (sequence) — sequence of `pydrex.minerals.Mineral`s
+    - `config` (dict) — PyDRex configuration dictionary
+    - `velocity_gradient` (array) — 3x3 velocity gradient matrix
+    - `time_steps` (array) ̒— time steps for the iteration loop
```
Patol75, line 47: Is there an extra character before the dash?

src/pydrex/core.py
```diff
@@ -1,49 +1,103 @@
...
+
+    Orientations and grain volume fractions are stored on the `Mineral`s.
+    Returns the finite strain ellipsoid after accumulated rotation.
```
Patol75, line 50: rotationS?

src/pydrex/core.py
```diff
@@ -1,49 +1,103 @@
...
+    """
+    finite_strain_ell = np.identity(3)  # Finite strain ellipsoid.
+    # Imposed macroscopic strain rate tensor.
+    strain_rate = (velocity_gradient + velocity_gradient.transpose()) / 2
+    # Strain rate scale (max. eigenvalue of strain rate).
+    strain_rate_max = np.abs(la.eigvalsh(strain_rate)).max()
+    # Dimensionless strain rate and velocity gradient.
+    nondim_strain_rate = strain_rate / strain_rate_max
+    nondim_velocity_gradient = velocity_gradient / strain_rate_max
+    # Reference resolved shear stress (relative slip system "softness")
```
Patol75, line 64: Delete?

Patol75 (changes requested):
Looks great, well done!

Patol75 (approved):

PR #30, opened 2022-03-31
Add CI file and migrate to scipy ivp

adigitoleo:
- Replace ad-hoc RK4 with scipy version
- clean up pytest fixtures
~~- add test file for aggregate tests~~
- add CI file for test automation

All tests should pass in ~10s currently. ~~The mineral aggregate test is a work-in-progress (it doesn't validate the results enough yet). I think there is still an issue with grain nucleation or something, because it looks like the grain size distribution is changed very drastically, very quicly. Some mockup plots of the olivine a-axis distribution also don't quite seem to match what they should right now.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: Continuous Integration
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  tests:
+    name: Unit tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install package
+        run: pip install -e .[test]
+
+      - name: Run tests
+        run: pytest
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -5,12 +5,14 @@ Acronyms:
     i.e. components of stress acting on each slip system in the grain reference frame
 
 """
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import IntEnum, unique
 
 import numpy as np
 import numba as nb
 from scipy.spatial.transform import Rotation
+from scipy.integrate import RK45
+from scipy import linalg as la
 
 import pydrex.core as _core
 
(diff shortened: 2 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,15 @@
...
+  pull_request:
+    branches:
+      - main
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: pip install -e .[test]
+      - run: pytest
```
Patol75, line 15: What about like this?

```
jobs:

  tests:
    name: Tests
    runs-on: ubuntu-latest

    steps:
      - name: Checkout repository
        uses: actions/checkout@v2

      - name: Install tests
        run: pip install -e .[test]

      - name: Run tests
        run: pytest
```

Patol75 (changes requested):

.github/workflows/ci.yml
adigitoleo: b6f87ef

Patol75:
`rebase` is always scary at first, no worries. Current changes look good to me, let me know if there are some further hurdles.

adigitoleo:
Pushing a squash rebase now, have removed the aggregate test for now, let's just get CI in because the aggregate test needs a little bit more work

src/pydrex/minerals.py
```diff
@@ -98,185 +100,155 @@ class Mineral:
...
-            orientations=self.orientations,
-            fractions=self.fractions,
-            strain_rate=strain_rate,
-            velocity_gradient=velocity_gradient,
-            stress_exponent=stress_exponent,
-            dislocation_exponent=dislocation_exponent,
-            nucleation_efficiency=nucleation_efficiency,
-            gbm_mobility=gbm_mobility,
-            volume_fraction=volume_fraction,
+        sol = RK45(
```
Patol75, line 194: Small comment here, we are using RK45 at the moment, but we may end up changing to another solver depending on future benchmarks/models.

src/pydrex/minerals.py
```diff
@@ -98,185 +100,155 @@ class Mineral:
...
-        self.orientations[mask, :, :] = self._orientations_init[mask, :, :]
-        self.fractions[mask] = gbs_threshold / self.n_grains
-        self.fractions /= self.fractions.sum()
+
+        while sol.status == "running":
+            sol.step()
+            finite_strain_ell, orientations, fractions = extract_vars(sol.y)
+            # Grain boundary sliding for small grains.
+            mask = fractions < config["gbs_threshold"] / self.n_grains
+            # No rotation: carry over previous orientations.
```
Patol75, line 210: Can you actually check if there is some mention of this somewhere in the DRex papers? Carrying over previous orientations sounds fair, but the Fortran reverts to the initial value, as we discussed.

Patol75 (approved):
Well done! I will merge, but please have a look at the two comments.

PR #31, opened 2022-04-11
Add more grain tests

adigitoleo:
Renaming some stuff, a new test at the end. Random initialisation for single grains is doing the right thing according to the (less strict) eigenvalue metric from Vollmer 1990.

Pushed to the wrong remote, will delete this branch after.


adigitoleo:
I should sort the eigenvalues before doing the comparison, one moment.

Patol75 (approved):

PR #32, opened 2022-04-11
test: Add very simple pure olivine A-type aggregate tests

adigitoleo:
Again using Vollmer 1990 eigenvalue diagnostic. The eigenvalue diagnostic is in fact converging for random initial orientations, so maybe I was plotting the wrong thing before.

EDIT: It's only converging for `n_grains=1` still...

However, it doesn't work when I do the same test at every timestep, only after the total integrated strain. Will investigate further.

### DIFF
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -219,6 +219,117 @@ class Mineral:
         self._fractions.append(fractions)
         return finite_strain_ell
 
+    def _update_orientations_drex2004(
+            self,
+            config,
+            deformation_gradient,
+            velocity_gradient,
+            n_iter,
+    ):
+        orientations = self._orientations[-1]
+        fractions = self._fractions[-1]
+        # Imposed macroscopic strain rate tensor.
+        strain_rate = (velocity_gradient + velocity_gradient.transpose()) / 2
+        # Strain rate scale (max. eigenvalue of strain rate).
+        strain_rate_max = np.abs(la.eigvalsh(strain_rate)).max()
+        # Dimensionless strain rate and velocity gradient.
+        nondim_strain_rate = strain_rate / strain_rate_max
+        nondim_velocity_gradient = velocity_gradient / strain_rate_max
+        # Volume fraction of the mineral phase.
+        if self.phase == MineralPhase.olivine:
+            volume_fraction = config["olivine_fraction"]
+        elif self.phase == MineralPhase.enstatite:
+            volume_fraction = config["enstatite_fraction"]
+        else:
+            assert False  # Should never happen.
+
+        # TODO: Improve this hack for choosing dt?
+        dt = 1.0 / n_iter / strain_rate_max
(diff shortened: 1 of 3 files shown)


adigitoleo:
Oops it's `[skip ci]` not `[skip-ci]`. All of the grain sizes become zero in the oblique shear test. Time to hunt for the problem with the grain size updates.

adigitoleo:
Nevermind, the `n_grains=1` test was only passing due to lucky numerical rounding error.

adigitoleo:
OK so single grain passes but aggregate with `n_grains=1` fails in exactly the same shear. Means that we need a maxstep for the ivp. I should allow it to be changed from within the Fluidity snippet, so people can put some value that depends on the mesh resolution. But we also want a default value, so I'm going to probably depend on `max_strain_rate` for that. Assuming mesh resolution of 1km then original PyDrex.py would have done `1 / 4 / velocity` so I'll try `0.25 * max_strain_rate` and go from there.

adigitoleo:
Can't get scipy to work, but if I use the original RK4 then it seems to work but is highly sensitive to the number and step size of the integration steps. Test is coming, for now with the old RK4 and a manual step tuning parameter.

What surprises me is that it fails again if the step size becomes too small, but that seems a bit weird for an integral...

adigitoleo:
Eigenvalue diagnostic for comparison:
pydrex_simple_shear_Y_single

adigitoleo:
Some updates:
- I don't think it makes sense to expect a single grain of random initial orientation to realign, CPO is a statistical property
- The reason why a random grain might not realign is because of N<5 slip systems (in fact N=3 for A-type) so dislocation creep cannot cause arbitrary realignment
- We need probably at least 1000 grains to do aggregate tests
- From my trials (original PyDRex.py), the orientation of 1000 or more grains will eventually converge to either 0 or +/-180 deg from the shear direction, MOST of the time
- Some of the time (~10%) it just doesn't ¯\_(ツ)_/¯ (numerical instability in advection scheme?)
- It can take a much longer time than I expected for convergence, I plot 8 x t* to be sure
- Before convergence, things look a lot more messy than fig 5 in Kaminski 2001, also I don't know how they got such a quick convergence
- Re-orthogonalising the orientation matrices after every iteration (cf. Drex-MATLAB, Fraters 2021) doesn't seem to improve things, unfortunately

There's also a lower priority issue:
- Sometimes it stabilises at 0, then flips to 180 later, I think due to grain size changes -- this would not be seen when plotting diagnostics like anisotropy strength (0 or 180 gives the same fast axis)

adigitoleo:
Update: closing as I will be pushing up reworked tests soon. There are enough differences to make it a new PR. The 0/180 angle issue is solved by applying the correct Bingham statistics. I can get consistent convergence for 2D simple shear models with initial orientations distributed in the 2D deformation plane. Fully 3D models will be a separate PR.

PR #33, opened 2022-04-11
test: Make assertion for random init case more robust

adigitoleo:
Quick fix for the assertion, have done the same for the other tests.


Patol75:
I have just realised that, but we probably want to use these [functions](https://numpy.org/doc/stable/reference/routines.testing.html).

Patol75 (approved):

PR #36, opened 2022-06-26
feat: Add bingham average and M-index diagnostics

adigitoleo:
Add methods to calculate bingham average and M-index (texture strength).
Will be used in other tests so if these fail other test results are invalid.
~~Not sure why the bingham averages for b- and c-axes are not orthogonal,
but that isn't important until we are looking at e.g. B-type olivine.~~ Fixed

I have also snuck the proper Vollmer diagnostics back in here, but they need more tests.

### DIFF
--- a/src/pydrex/diagnostics.py
+++ b/src/pydrex/diagnostics.py
@@ -7,55 +7,45 @@ import scipy.linalg as la
 import pydrex.tensors as _tensors
 
 
-def bingham_average(orientations):
-    """Compute bingham averages from olivine orientation matrices.
+def Bingham_average(orientations, axis="a"):
+    """Compute Bingham averages from olivine orientation matrices.
 
-    Returns a tuple of vectors which are the averaged
-    a-, b- and c-axes orientations, respectively.
+    Returns the antipodally symmetric average orientation
+    of the given crystallographic `axis`, or the a-axis by default.
+    Valid axis specifiers are "a" for [100], "b" for [010] and "c" for [001].
+
+    See e.g. [Watson 1966].
+
+    [Watson 1966]: https://doi.org/10.1086%2F627211
 
     """
+    if axis == "a":
+        row = 0
+    elif axis == "b":
+        row = 1
+    elif axis == "c":
+        row = 2
+    else:
+        raise ValueError(f"axis must be 'a', 'b', or 'c', not {axis}")
+
+    # Scatter (inertia) matrix, see eq. 2.4 in Watson 1966.
+    scatter = np.zeros((3, 3))
+    scatter[0, 0] = np.sum(orientations[:, row, 0] ** 2)
+    scatter[1, 1] = np.sum(orientations[:, row, 1] ** 2)
+    scatter[2, 2] = np.sum(orientations[:, row, 2] ** 2)
+    scatter[1, 0] = np.sum(orientations[:, row, 0] * orientations[:, row, 1])
+    scatter[2, 0] = np.sum(orientations[:, row, 0] * orientations[:, row, 2])
+    scatter[2, 1] = np.sum(orientations[:, row, 1] * orientations[:, row, 2])
+
     # https://courses.eas.ualberta.ca/eas421/lecturepages/orientation.html
     # Eigenvector corresponding to largest eigenvalue is the mean direction.
     # SciPy returns eigenvalues in ascending order (same order for vectors).
-    a_mean = np.zeros((3, 3))
-    a_mean[0, 0] = np.sum(orientations[:, 0, 0] ** 2)
-    a_mean[1, 1] = np.sum(orientations[:, 0, 1] ** 2)
-    a_mean[2, 2] = np.sum(orientations[:, 0, 2] ** 2)
-    a_mean[0, 1] = np.sum(orientations[:, 0, 0] * orientations[:, 0, 1])
-    a_mean[0, 2] = np.sum(orientations[:, 0, 0] * orientations[:, 0, 2])
-    a_mean[1, 2] = np.sum(orientations[:, 0, 1] * orientations[:, 0, 2])
-    _, a_eigenvectors = la.eigh(_tensors.upper_tri_to_symmetric(a_mean))
-    a_vector = a_eigenvectors[:, -1]
-
-    b_mean = np.zeros((3, 3))
-    b_mean[0, 0] = np.sum(orientations[:, 1, 0] ** 2)
-    b_mean[1, 1] = np.sum(orientations[:, 1, 1] ** 2)
-    b_mean[1, 2] = np.sum(orientations[:, 1, 2] ** 2)
-    b_mean[0, 1] = np.sum(orientations[:, 1, 0] * orientations[:, 1, 1])
-    b_mean[0, 2] = np.sum(orientations[:, 1, 0] * orientations[:, 1, 2])
-    b_mean[1, 2] = np.sum(orientations[:, 1, 1] * orientations[:, 1, 2])
-    _, b_eigenvectors = la.eigh(_tensors.upper_tri_to_symmetric(b_mean))
-    b_vector = b_eigenvectors[:, -1]
-
-    c_mean = np.zeros((3, 3))
-    c_mean[0, 0] = np.sum(orientations[:, 2, 0] ** 2)
-    c_mean[1, 1] = np.sum(orientations[:, 2, 1] ** 2)
-    c_mean[2, 2] = np.sum(orientations[:, 2, 2] ** 2)
-    c_mean[0, 1] = np.sum(orientations[:, 2, 0] * orientations[:, 2, 1])
-    c_mean[0, 2] = np.sum(orientations[:, 2, 0] * orientations[:, 2, 2])
-    c_mean[1, 2] = np.sum(orientations[:, 2, 1] * orientations[:, 2, 2])
-    _, c_eigenvectors = la.eigh(_tensors.upper_tri_to_symmetric(c_mean))
-    c_vector = c_eigenvectors[:, -1]
-
-    return (
-        a_vector / la.norm(a_vector),
-        b_vector / la.norm(b_vector),
-        c_vector / la.norm(c_vector),
-    )
+    mean_vector = la.eigh(scatter)[1][:, -1]
+    return mean_vector / la.norm(mean_vector)
 
 
 def resample_orientations(orientations, fractions, n_samples=None):
-    """Generate samples from `orientations` based on volume distribution.
+    """Generate new samples from `orientations` based on volume distribution.
 
     If the optional number of samples is not specified,
     it will be set to the number of original "grains" (length of `fractions`).
@@ -63,13 +53,14 @@ def resample_orientations(orientations, fractions, n_samples=None):
     """
     if n_samples is None:
         n_samples = len(fractions)
-    sort_ascending = np.argsort(fractions)[::-1]
-    frac_ascending = np.take(fractions, sort_ascending)
+    # sort_ascending = np.argsort(fractions)[::-1]
+    # frac_ascending = np.take(fractions, sort_ascending)
     # Cumulative volume fractions.
-    cumfrac = np.cumsum(frac_ascending)
+    cumfrac = fractions.sort().cumsum()
     # Number of new samples with volume less than each cumulative fraction.
-    count_less = [np.searchsorted(cumfrac, r) for r in rn.random_sample(n_samples)]
-    return np.stack([orientations[i] for i in count_less])
+    rng = rn.default_rng()
+    count_less = np.searchsorted(cumfrac, rng.random(n_samples))
+    return orientations[count_less]
 
 
 def M_index(orientations):
@@ -80,19 +71,23 @@ def M_index(orientations):
     [Skemer et al. 2005]: https://doi.org/10.1016/j.tecto.2005.08.023
 
     """
-    misorientations = np.fromiter(
-        (misorientation_angle(A, B) for A, B in it.combinations(orientations, 2)),
-        dtype=float,
-    )
+    misorientations = [
+        misorientation_angle(A, B) for A, B in it.combinations(orientations, 2)
+    ]
     # Number of misorientations within 1° bins.
-    count_misorientations, _ = np.histogram(misorientations, list(range(121)))
-    return (1 / 2 / len(misorientations)) * np.sum(
-        np.abs(
-            [
-                misorientations_random(i, i + 1) * len(misorientations)
-                for i in range(120)
-            ]
-            - count_misorientations
+    count_misorientations, _ = np.histogram(misorientations, bins=120, range=(0, 120))
+    return (
+        1
+        / 2
+        / len(misorientations)
+        * np.sum(
+            np.abs(
+                [
+                    misorientations_random(i, i + 1) * len(misorientations)
+                    - count_misorientations[i]
+                    for i in range(120)
+                ]
+            )
         )
     )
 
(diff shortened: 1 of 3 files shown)


src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,193 @@
...
+
+    See [Skemer et al. 2005].
+
+    [Skemer et al. 2005]: https://doi.org/10.1016/j.tecto.2005.08.023
+
+    """
+    misorientations = np.fromiter(
+        (misorientation_angle(A, B) for A, B in it.combinations(orientations, 2)),
+        dtype=float,
+    )
```
Patol75, line 86: Any reason not to use
`misorientations = [misorientation_angle(A, B) for A, B in it.combinations(orientations, 2)]`
?

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,193 @@
...
+    # Number of misorientations within 1° bins.
+    count_misorientations, _ = np.histogram(misorientations, list(range(121)))
```
Patol75, line 88: `count_misorientations, _ = np.histogram(misorientations, bins=120, range=(0, 120))`

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,193 @@
...
+    return (1 / 2 / len(misorientations)) * np.sum(
```
Patol75, line 89: No need for the extra parentheses.

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,193 @@
...
+        np.abs(
+            [
+                misorientations_random(i, i + 1) * len(misorientations)
+                for i in range(120)
+            ]
+            - count_misorientations
```
Patol75, line 95: Could perhaps put the `- count_misorientations` within the list comprehension.

tests/test_diagnostics.py
```diff
@@ -0,0 +1,102 @@
...
+from scipy.spatial.transform import Rotation
+
+from pydrex import diagnostics as _diagnostics
+
+
+class TestBinghamStats:
+    """Tests for antipodally symmetric (Bingham) statistics."""
+
+    def test_average_0(self):
+        """Test bingham average of vectors aligned to the reference frame."""
```
Patol75, line 13: Bingham

tests/test_diagnostics.py
```diff
@@ -0,0 +1,102 @@
...
+                [0, 0, np.pi / 2],
+            ]
+        ).as_matrix()
+        a_mean, b_mean, c_mean = _diagnostics.bingham_average(orientations)
+        assert np.array_equal(a_mean, [0, 1, 0])
+        assert np.array_equal(b_mean, [1, 0, 0])
+        assert np.array_equal(c_mean, [0, 0, 1])
+
+    def test_average_spread10X(self):
+        """Test bingham average of vectors spread within 10° of the ±X-axis."""
```
Patol75, line 34: Bingham

tests/test_diagnostics.py
```diff
@@ -0,0 +1,102 @@
...
+        orientations = Rotation.from_rotvec(
+            np.stack(
+                [
+                    [0, x * np.pi / 18 - np.pi / 36, x * np.pi / 18 - np.pi / 36]
+                    for x in rn.random_sample(100)
```
Patol75, line 39: `for x in rn.default_rng().random(100)`

Patol75 (changes requested):
Great stuff putting this whole thing together! Quite a few comments on my side, please have a look. :)

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,193 @@
...
+    c_vector = c_eigenvectors[:, -1]
+
+    return (
+        a_vector / la.norm(a_vector),
+        b_vector / la.norm(b_vector),
+        c_vector / la.norm(c_vector),
+    )
+
+
+def resample_orientations(orientations, fractions, n_samples=None):
```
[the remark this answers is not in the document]
adigitoleo: This is an implementation of what Fraters described in the supplementary information section 2:

> The weighted data is the result of a procedure which effectively resamples the aggregate of N grains, by sampling the largest grains a number of times that is proportional to the volume fraction of the grain and leaving the smallest grains unrepresented in the weighted sample. In this procedure the grains are sorted by volume fraction from smallest to largest. The cumulative volume fraction is then calculated
for each grain. Next, a vector of N random numbers with values between 0 and 1 is created. For each random number we determine how many grains have a cumulative volume fraction less than this random number. This number is then used as the index into the sorted list of grains, and the value of the volume fraction and Euler angle for this index is stored. This is repeated for all N random numbers. This procedure reproduces a volumetric random sampling of the aggregate.

If I understand correctly, we basically resample from [the empirical distribution](https://en.wikipedia.org/wiki/Empirical_distribution_function). Therefore upsampling/downsampling just works, because for each sample we just pick a value from the original data based on the weights and return it. I will add tests though for `n_samples != len(fractions)`.

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,193 @@
...
+
+    """
+    M, N = symmetry
+    assert low >= 0
+    if (M == 2 and N == 4) or (M == 3 and N == 6):
+        maxval = 120
+    elif (M == 4 and N == 8) or (M == 6 and N == 12):
+        maxval = 90
+    else:
+        maxval = 180
```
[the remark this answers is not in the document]
adigitoleo: Oh! Haha, yeah why not, I'll bump the python version in CI as well then.

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,193 @@
...
+    # SciPy returns eigenvalues in ascending order (same order for vectors).
+    a_mean = np.zeros((3, 3))
+    a_mean[0, 0] = np.sum(orientations[:, 0, 0] ** 2)
+    a_mean[1, 1] = np.sum(orientations[:, 0, 1] ** 2)
+    a_mean[2, 2] = np.sum(orientations[:, 0, 2] ** 2)
+    a_mean[0, 1] = np.sum(orientations[:, 0, 0] * orientations[:, 0, 1])
+    a_mean[0, 2] = np.sum(orientations[:, 0, 0] * orientations[:, 0, 2])
+    a_mean[1, 2] = np.sum(orientations[:, 0, 1] * orientations[:, 0, 2])
+    _, a_eigenvectors = la.eigh(_tensors.upper_tri_to_symmetric(a_mean))
+    a_vector = a_eigenvectors[:, -1]
```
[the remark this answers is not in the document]
adigitoleo: It's not the same shape though. I guess I can do `la.eigh(_tensors.upper_tri_to_symmetric(scatter))[1][:, -1]` ?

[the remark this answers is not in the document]
Patol75: The line I wrote should give you the eigenvector corresponding with the highest eigenvalue of a 3x3 matrix.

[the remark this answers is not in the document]
adigitoleo: I like to have it in flat array format:

```
In [12]: la.eigh(np.eye(3))[1][:, -1]
Out[12]: array([0., 0., 1.])

In [13]: la.eigh(np.eye(3), subset_by_index=[2, 2])[1]
Out[13]:
array([[0.],
       [0.],
       [1.]])

In [14]: la.eigh(np.eye(3), subset_by_index=[2, 2])[1].T[0]
Out[14]: array([0., 0., 1.])
```

Is there a clever way to do that using the kwargs? I think the eigenvectors will always be ordered by increasing eigenvalues anyway right?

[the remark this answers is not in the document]
Patol75: What about
`a_vector = la.eigh(a_mean.T, subset_by_index=[2, 2])[1].T`
?
I have just discovered that, by default, `eigh` only reads the lower triangle part of the supplied array.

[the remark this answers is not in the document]
Patol75: What about
`a_vector = la.eigh(a_mean.T, subset_by_index=[2, 2])[1].T`
?
I have just discovered that, by default, `eigh` only reads the lower triangle part of the supplied array.

adigitoleo:
Thanks for the comments. I will add one more commit with some extra tests for orientation data resampling soon.

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,190 @@
...
+    if n_samples is None:
+        n_samples = len(fractions)
+    # sort_ascending = np.argsort(fractions)[::-1]
+    # frac_ascending = np.take(fractions, sort_ascending)
+    # Cumulative volume fractions.
+    cumfrac = fractions.sort().cumsum()
+    # Number of new samples with volume less than each cumulative fraction.
+    rng = rn.default_rng()
+    count_less = np.searchsorted(cumfrac, rng.random(n_samples))
+    return orientations[count_less]
```
Patol75, line 63: Hmm, wait here. Based on what you explained above, should not we actually bring back `argsort` and then write the following:
```
def resample_orientations(orientations, fractions, n_samples=None):
    """Generate new samples from `orientations` based on volume distribution.

    If the optional number of samples is not specified,
    it will be set to the number of original "grains" (length of `fractions`).
    """
    if n_samples is None:
        n_samples = len(fractions)
    sort_ascending = np.argsort(fractions)
    # Cumulative volume fractions.
    cumfrac = fractions[sort_ascending].cumsum()
    # Number of new samples with volume less than each cumulative fraction.
    rng = rn.default_rng()
    count_less = np.searchsorted(cumfrac, rng.random(n_samples))
    return orientations[sort_ascending][count_less]
```
? Also, don't you also need to return the associated volume fractions? It is mentioned in your explanation.

adigitoleo: Yeah I think that is correct. I really need some better tests for this resampling business at some point (cough Pole Figures cough). I've also added volume fractions in the return, https://github.com/Patol75/PyDRex/commit/b03ba3742286f68fbb662c5b1ff6c5792011ba3d#diff-e6b4c1fadcae25149a1eb99ee1934f02ef26094e179e218dc2116af77edce5f7R102-R108

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,230 @@
...
+    """
+    if n_samples is None:
+        n_samples = len(fractions)
+    sort_ascending = np.argsort(fractions)
+    # Cumulative volume fractions
+    cumfrac = fractions[sort_ascending].cumsum()
+    # Number of new samples with volume less than each cumulative fraction.
+    rng = rn.default_rng()
+    count_less = np.searchsorted(cumfrac, rng.random(n_samples))
+    return orientations[count_less], fractions[count_less]
```
Patol75, line 108: So you don't need to sort `orientations` before you apply `count_less`?

adigitoleo: Woops, fixed. thanks

adigitoleo:
OK had to fiddle with the symmetry diagnostics tolerances but hopefully it is not too flaky now. Thanks for the attention to the resampling function, this was something I hadn't used much yet.

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,231 @@
+import itertools as it
+
+import numpy as np
+from numpy import random as rn
+import scipy.linalg as la
+
+import pydrex.tensors as _tensors
```
Patol75, line 7: Not used anymore.

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def Bingham_average(orientations, axis="a"):
```
Patol75, line 10: Probably best to keep the function name lower-cased (my previous comment was mainly referring to the docstring, even though I could have made it clearer).

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,231 @@
...
+    for ternary texture classification.
+    Returns the tuple (P, G, R) where
+
+    $$
+    P = (λ_{1} - λ_{2}) / N
+    G = 2 (λ_{2} - λ_{3}) / N
+    R = 3 λ_{3} / N
+    $$
+
+    for N the sum of the eigenvalues descending eigenvalues $λ_{i}$
```
Patol75, line 52: `with N the sum of the eigenvalues` ?

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+    match symmetry:
+        case (2, 4) | (3, 6):
+            maxval = 120
+        case (4, 8) | (6, 12):
+            maxval = 90
+        case (1, 1) | (2, 2):
+            maxval = 180
+        case _:
+            raise ValueError(f"incorrect symmetry values (M,N) = {symmetry}")
```
Patol75, line 177: Possibly missing a space between `M,` and `N`.

src/pydrex/diagnostics.py
```diff
@@ -0,0 +1,231 @@
...
+                    - 2
+                    * np.sin(d)
+                    * (
+                        2 * np.rad2deg(np.arccos(a / np.sqrt(ν - 1)))
+                        + a * np.rad2deg(np.arccos(1 / np.sqrt(ν - a**2)))
+                    )
+                )
+            )
+        else:
+            assert False  # Should never happen.
```
Patol75, line 229: Maybe `raise ValueError("message")` to make it clearer.

tests/test_diagnostics.py
```diff
@@ -0,0 +1,166 @@
...
+        orientations = Rotation.random(1000).as_matrix()
+        assert np.isclose(_diagnostics.M_index(orientations), 0.44, atol=1e-2)
+
+    def test_texture_spread10X(self):
+        """Test M-index for grains spread within 10° of the ±X axis."""
+        orientations = Rotation.from_rotvec(
+            np.stack(
+                [
+                    [0, x * np.pi / 18 - np.pi / 36, x * np.pi / 18 - np.pi / 36]
+                    for x in rn.random_sample(100)
```
Patol75, line 128: Oh, another nasty `random_sample`!

tests/test_diagnostics.py
```diff
@@ -0,0 +1,166 @@
...
+        ).as_matrix()
+        assert np.isclose(_diagnostics.M_index(orientations), 0.99, atol=1e-2)
+
+    def test_texture_spread30X(self):
+        """Test M-index for grains spread within 45° of the ±X axis."""
+        orientations = Rotation.from_rotvec(
+            np.stack(
+                [
+                    [0, x * np.pi / 4 - np.pi / 8, x * np.pi / 4 - np.pi / 8]
+                    for x in rn.random_sample(100)
```
Patol75, line 140: And another one!

tests/test_diagnostics.py
```diff
@@ -0,0 +1,166 @@
...
+        for i, factor in enumerate(factors):
+            orientations = Rotation.from_rotvec(
+                np.stack(
+                    [
+                        [
+                            0,
+                            x * np.pi / factor - np.pi / factor / 2,
+                            x * np.pi / factor - np.pi / factor / 2,
+                        ]
+                        for x in rn.random_sample(500)
```
Patol75, line 159: That has to be the last one. :)

Patol75 (approved):
Just some additional, very minor comments. Looking great!

tests/test_diagnostics.py
```diff
@@ -0,0 +1,166 @@
...
+        """Test M-index for grains spread within 45° of the ±X axis."""
+        orientations = Rotation.from_rotvec(
+            np.stack(
+                [
+                    [0, x * np.pi / 4 - np.pi / 8, x * np.pi / 4 - np.pi / 8]
+                    for x in rn.random_sample(100)
+                ]
+            )
+        ).as_matrix()
+        assert np.isclose(_diagnostics.M_index(orientations), 0.84, atol=0.4e-1)
```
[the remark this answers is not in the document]
adigitoleo: Ah yes I mentioned in #33 that I didn't see the point, but actually they do have `err_msg` argument so I don't have any problem with changing to this. I can't see any `assert_isclose` though.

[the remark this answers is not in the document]
Patol75:
```
>>> assert np.allclose(np.arange(3), [0, 1, 3])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
>>> np.testing.assert_allclose(np.arange(3), [0, 1, 3])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/thomas/.local/lib/python3.10/site-packages/numpy/testing/_private/utils.py", line 1530, in assert_allclose
    assert_array_compare(compare, actual, desired, err_msg=str(err_msg),
  File "/home/thomas/.local/lib/python3.10/site-packages/numpy/testing/_private/utils.py", line 844, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

Mismatched elements: 1 / 3 (33.3%)
Max absolute difference: 1
Max relative difference: 0.33333333
 x: array([0, 1, 2])
 y: array([0, 1, 3])
```

That is the sort of difference you get.

src/pydrex/diagnostics.py
adigitoleo: I should listen to my linter :joy:

Patol75 (approved):
Thanks. The new test is failing though.

adigitoleo:
I'll loosen the tolerances a bit.

PR #38, opened 2022-07-11
Simple shear test

adigitoleo:
Adds a 2D simple shear test. Optionally save the output figure with:

    pytest --outdir="out/"

where "out/" is the name of an existing directory (can be "." for cwd).

Have been tweaking the tolerances for a while, hopefully not too flaky... M* is grain boundary mobility and the red dashed line is the analytical result for a single grain if it started at 45°.

image


Patol75 (changes requested):
Some early comments.

Patol75 (approved):

Patol75 (approved):

PR #40, opened 2022-07-13
GBS test and small test fixes

adigitoleo:
Adds a 2d simple shear test which varies the grain boundary sliding parameter.
This uses parameters from Kaminski 2004. Grain boundary sliding acts as an
upper bound on texture strength and a lower bound on CPO alignment.
The initial contition is set up for shear along a different axis,
but slip on the same slip system as the GBM test from https://github.com/Patol75/PyDRex/pull/38
(this tests that the model is working regardless of the orientation of the 2D shear).

image


Patol75 (approved):
Looks great!

PR #45, opened 2022-08-02
Fix flaky tests

adigitoleo:
Fixes flaky tests discovered in #44 .


adigitoleo:
Now I understand why numerical people don't like the word "random".

PR #46, opened 2022-08-19
Pathline fixes and vtk reader test

adigitoleo:
Switch to RBF for interpolation and fix vtk readers + simple regression test.

Benchmark to test pathlines is also on the way, but currently taking a long time to run...

I only plan on uploading maybe one 3D vtu file (and maybe one simple pvtu folder) for other tests, so that the repo doesn't get too bloated. For most larger benchmarks I will only upload the .flml, they won't be automated anyway and will go into `examples` instead of `tests`.


src/pydrex/vtk_helpers.py
```diff
@@ -65,33 +43,39 @@ def read_tuple_array(points, fieldname, skip_z=False):
...
-    a = np.array([data.GetTuple(i) for i in range(data.GetNumberOfTuples())])
-    if skip_z:
-        return _skip_column_every(a, 3)
-    return a
+    values = np.array([data.GetTuple(i) for i in range(data.GetNumberOfTuples())])
+
+    if values.shape[1] == 9:  # Matrix data.
+        values = values.reshape((values.shape[0], 3, 3))
+        if skip3:
+            return values[:, 0:2, 0:2]
```
Patol75, line 51: Zeros are not required before the columns.

src/pydrex/vtk_helpers.py
```diff
@@ -65,33 +43,39 @@ def read_tuple_array(points, fieldname, skip_z=False):
...
+        return values
 
+    if skip3:
+        return values[:, 0:2]
```
Patol75, line 55: Same here.

Patol75 (approved):
Looks good to me. Remember to make the necessary changes for the time- and/or space-varying velocity (gradient) fields.

adigitoleo:
Will merge this first, the IVP fixes will be coming from a different branch.

PR #50, opened 2022-09-07
Cornerflow test

adigitoleo:
Cornerflow test, integration fixes and logging.

Draft because I still need to:

- [x] ~~Optimise time to create `Mineral`s, which seems a bit slow compared to the solver time~~ Nope, it's the diagnostics and output happening after one Mineral finishes and before the next one is created, not optimising this here
- [x] ~~Debug why my CPO is not aligning to the extent shown in Kaminski 2002 fig. 5~~ Postponed until further analysis, could be differences in the spherical averaging.
- [x] Stop dumping so much data in logs
- [x] Finish the second test with the (analytical) pathline
- [x] Add actual asserts so the tests are not just toy examples :)
- [x] ~~Set up a test ordering so more complex tests are run later, and quick tests can fail early and give a big red X~~ Nope, there's not an easy way to do it in pytest and I don't want too many hacks in my testing code
- [x] Figure out why pytest is not respecting `--log-cli-level` in CI -- various reasons, including buggy interaction of pytest CLI args and other loggers (https://github.com/pytest-dev/pytest/issues/10283)


adigitoleo:
This is fun: pytest is behaving in three different ways in three copies of the same environment (Arch, Ubuntu, CI).

Locally, I can't reproduce the "always verbose" log output which is crashing CI. I'm going to have to try setting some options blind, see if it stops.

adigitoleo:
Example output for 2cm/yr
image

Angle only reaches ~70° compared to much shallower angle in Kaminski 2002 fig 5, using same GBM and nucleation parameters. Not sure what kind of velocity boundary was imposed for their model though.

src/pydrex/minerals.py
```diff
@@ -96,13 +100,14 @@ class Mineral:
 
     A `Mineral` stores CPO data for an aggregate of grains*.
     Additionally, mineral fabric type and deformation regime
```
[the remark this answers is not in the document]
adigitoleo: The github UI is not helping me very much here, are you referring to L102 or L103? If you mean L102, then no that's fine, we are still tracking the mineral fabric, however there is no `pydrex.fabric` module, I removed it some time back but didn't update the docstring.

adigitoleo:
I'll need to revisit this later to fully understand what is going on, but for now I think it's worth having these even just to check that the pathline routines run properly. This PR also got a bit out of hand so it includes a bunch of infrastructure that will be useful later (logging, visualisation, etc.)

src/pydrex/logger.py
```diff
@@ -0,0 +1,100 @@
...
+            logging.ERROR: self.colorfmt("31"),
+            logging.INFO: self.colorfmt("32"),
+            logging.WARNING: self.colorfmt("33"),
+            logging.DEBUG: self.colorfmt("34"),
+        }
+        self._style._fmt = format_specs.get(record.levelno)
+        return super().format(record)
+
+
+# To create a new logger we use getLogger as recommeded by the logging docs.
```
Patol75, line 39: recommended

tests/conftest.py
```diff
@@ -1,20 +1,67 @@
...
+from _pytest.logging import LoggingPlugin, _LiveLoggingStreamHandler
+from numpy import random as rn
+
+from pydrex import logger as _log
+
+matplotlib.use("Agg")  # Stop matplotlib from looking for $DISPLAY in env.
+_log.quiet_aliens()  # Stop imported modules from spamming the logs.
+
+
+# The default pytest logging-plugin always creates its own handlers...
```
Patol75, line 15: logging plugin

Patol75 (commented):
Great work, well done!

Patol75 (approved):

PR #51, opened 2022-09-09
ENH: Update pre-commit and repository

Patol75:
Just a quick update, tidying things around.

@adigitoleo Let me know if it looks fine to you.


.pre-commit-config.yaml
```diff
@@ -31,13 +31,23 @@ repos:
...
+        args: ["--profile", "black", "--filter-files"]
 -   repo: https://github.com/asottile/pyupgrade
     rev: v2.37.3
     hooks:
     -   id: pyupgrade
+-   repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
+    hooks:
+    -   id: flake8
+        args: ["--max-line-length", "88", "--extend-ignore", "E203,E722"]
```
adigitoleo, line 47: Good point, I have similar settings in my editor. Can we also ignore W503, which violates PEP8 and is also possibly scheduled for deprecation: https://www.flake8rules.com/rules/W503.html

tests/conftest.py
```diff
@@ -41,7 +39,7 @@ def params_Kaminski2001_fig5_solid():
         "gbm_mobility": 0,
         "gbs_threshold": 0,
         "nucleation_efficiency": 5,
-        "minerals": ("olivine"),
```
adigitoleo, line 44: I think I wanted for these to be a tuple so the type is predictable, but for that it should be `("olivine",)` with the comma, which I forgot about.

adigitoleo (commented):
Thanks for cleaning things up, I should go back over `run.py` and the `solve_*` methods in `core` at some point. Two small things which I can fix after lunch if you haven't done so already.

adigitoleo (approved):

PR #66, opened 2023-01-28
feat: Add script to read fluidity h5part files

adigitoleo:
h5part -> NPZ, a bit slow to convert but gives a 60x smaller file, use something like the following to animate pole figures along a pathline:

```
import numpy as np
from scipy import linalg as la
from matplotlib import pyplot as plt
from matplotlib.animation import FuncAnimation

from pydrex import diagnostics as _diagnostics
from pydrex import minerals as _minerals


def set_polefig_axis(ax, ref_axes="xz"):
    ax.set_axis_off()
    ax.set_xlim((-1.1, 1.1))
    ax.set_ylim((-1.1, 1.1))
    ax.set_aspect("equal")
    _circle_points = np.linspace(0, np.pi * 2, 100)
    ax.plot(np.cos(_circle_points), np.sin(_circle_points), linewidth=1, color="k")
    ax.axhline(0, color="k")
    ax.text(1.05, 0.5, ref_axes[0], verticalalignment="center", transform=ax.transAxes)
    ax.axvline(0, color="k")
    ax.text(
        0.5, 1.05, ref_axes[1], horizontalalignment="center", transform=ax.transAxes
    )


def poles(orientations, ref_axes="xz", hkl=[1, 0, 0]):
    upward_axes = next((set("xyz") - set(ref_axes)).__iter__())
    axes_map = {"x": 0, "y": 1, "z": 2}
    directions = np.tensordot(orientations.transpose([0, 2, 1]), hkl, axes=(2, 0))
    directions_norm = la.norm(directions, axis=1)
    directions[:, 0] /= directions_norm
    directions[:, 1] /= directions_norm
    directions[:, 2] /= directions_norm
    stereograph_xvals = directions[:, axes_map[ref_axes[0]]] / (
        1 + np.abs(directions[:, axes_map[upward_axes]])
    )
    stereograph_yvals = directions[:, axes_map[ref_axes[1]]] / (
        1 + np.abs(directions[:, axes_map[upward_axes]])
    )
    return stereograph_xvals, stereograph_yvals


mineral = _minerals.Mineral.from_file("pydrex_RK45_once.npz")
fig = plt.figure()
ax = fig.add_subplot(111)
orientations_resampled = [
    _diagnostics.resample_orientations(_orientations, _fractions)[0]
    for _orientations, _fractions in zip(mineral.orientations, mineral.fractions)
]
# xvals, yvals = poles(orientations_resampled[0])
xvals, yvals = poles(mineral.orientations[0])
pf = ax.scatter(xvals, yvals, marker=".", alpha=0.33)


def _init():
    set_polefig_axis(ax)
    return (pf,)


def _update(frame):
    # xvals, yvals = poles(orientations_resampled[frame + 1])
    xvals, yvals = poles(mineral.orientations[frame + 1])
    pf.set_offsets(np.array(list(zip(xvals, yvals))))
    return (pf,)


animation = FuncAnimation(
    fig,
    _update,
    frames=range(len(orientations_resampled) - 1),
    init_func=_init,
    blit=True,
)

animation.save("temp_pole_fig.mp4")
```

### DIFF
--- a/tools/h5part_reader.py
+++ b/tools/h5part_reader.py
@@ -1,0 +1,134 @@
+#!/bin/env python
+"""Script to read CPO data from fluidity h5part files.
+
+CPO data is saved to a NPZ file which can be read by `pydrex.mineral.Mineral.from_file`.
+The data is extracted from particles with a `scalar_attribute_array` named 'CPO',
+and saved to the NPZ file under fields named with the postfix set to the particle ID.
+For example, the fields 'meta_1', 'fractions_1' and 'orientations_1' store data for the
+particle with id 1. The particle positions are stored in the additional fields `x`, `y`
+and `z`.
+
+"""
+import argparse
+import io
+import os
+from zipfile import ZipFile
+
+import h5py
+import numpy as np
+
+from pydrex import minerals as _minerals
+
+
+def _get_args() -> argparse.Namespace:
+    description, epilog = __doc__.split(os.linesep + os.linesep, 1)
+    parser = argparse.ArgumentParser(description=description, epilog=epilog)
+    parser.add_argument(
+        "input",
+        help="input file (.h5part)",
+    )
+    parser.add_argument(
+        "-p",
+        "--phase",
+        help="mineral phase, default: olivine",
+        default="olivine",
+    )
+    parser.add_argument(
+        "-f",
+        "--fabric",
+        help="olivine fabric type, default: A, use 'N' for enstatite",
+        default="A",
+    )
+    parser.add_argument(
+        "-o",
+        "--output",
+        help="filename for the output NPZ file",
+    )
+    parser.add_argument(
+        "-n",
+        "--ngrains",
+        type=int,
+        help="number of grains in the polycrystal",
+        required=True,
+    )
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    args = _get_args()
+
+    if not args.input.endswith(".h5part"):
+        raise ValueError(
+            f"expected input file with .h5part extension, not {args.input}"
+        )
+    outfile = args.input[:-6] + "npz"  # Replace `.h5part` with `.npz`.
+    if type(args.output) == str and args.output != "":
+        if args.output.endswith(".npz"):
+            outfile = args.output
+        else:
+            raise ValueError(f"can only save to NPZ format, not {args.output}")
+
+    with h5py.File(args.input) as file:
+        # Don't use the last timestep,
+        # fluidity deletes the detector and just writes empty data there.
+        n_timesteps = len(file.keys()) - 1
+
+        for particle_id in file["Step#0/id"][:]:
+            option_map = {
+                "olivine": _minerals.MineralPhase.olivine,
+                "enstatite": _minerals.MineralPhase.enstatite,
+                "A": _minerals.OlivineFabric.A,
+                "B": _minerals.OlivineFabric.B,
+                "C": _minerals.OlivineFabric.C,
+                "D": _minerals.OlivineFabric.D,
+                "E": _minerals.OlivineFabric.E,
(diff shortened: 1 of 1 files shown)


tools/h5part_reader.py
```diff
@@ -0,0 +1,134 @@
...
+    if not args.input.endswith(".h5part"):
+        raise ValueError(
+            f"expected input file with .h5part extension, not {args.input}"
+        )
+    outfile = args.input[:-6] + "npz"  # Replace `.h5part` with `.npz`.
+    if type(args.output) == str and args.output != "":
+        if args.output.endswith(".npz"):
+            outfile = args.output
+        else:
+            raise ValueError(f"can only save to NPZ format, not {args.output}")
```
Patol75, line 69: Here, you first define `outfile`, but then you only proceed if two conditions are met, in which case you define `outfile` again.

tools/h5part_reader.py
```diff
@@ -0,0 +1,134 @@
...
+
+    with h5py.File(args.input) as file:
+        # Don't use the last timestep,
+        # fluidity deletes the detector and just writes empty data there.
+        n_timesteps = len(file.keys()) - 1
```
Patol75, line 74: That may prove to be unwanted for other simulations in the future.

tools/h5part_reader.py
```diff
@@ -0,0 +1,134 @@
...
+            mineral.fractions = _fractions
+            mineral.orientations = _orientations
+            mineral.save(outfile, postfix=_postfix)
+            archive = ZipFile(outfile, mode="a", allowZip64=True)
+            for key, data in zip(("x", "y", "z"), (x, y, z)):
+                with archive.open(f"{key}_{_postfix}", "w", force_zip64=True) as file:
+                    buffer = io.BytesIO()
+                    np.save(buffer, data)
+                    file.write(buffer.getvalue())
+                    buffer.close()
```
Patol75, line 134: Why not use [`numpy.savez`](https://numpy.org/doc/stable/reference/generated/numpy.savez.html)?

tools/h5part_reader.py
```diff
@@ -0,0 +1,134 @@
...
+
+                # Extract particle position.
+                x[t] = file[f"Step#{t}/x"][particle_id - 1]
+                y[t] = file[f"Step#{t}/y"][particle_id - 1]
+                z[t] = file[f"Step#{t}/z"][particle_id - 1]
+
+                # Extract CPO data.
+                vals = np.empty(args.ngrains * 10)
+                for n in range(len(vals)):
+                    vals[n] = file[f"Step#{t}/CPO{n+1}"][particle_id - 1]
```
Patol75, line 105: Am I understanding correctly that you have many `CPO` particle attributes?

Patol75 (changes requested):
A few things to address, but looking good overall.

tools/h5part_reader.py
adigitoleo: This gives the option to use `-o <outfile>` to explicitly name the NPZ output file, otherwise it just uses the input file name but with the extension changed to `.npz`. Maybe there's a cleaner way but that's why I'm redefining: default is <input_file>.npz but it can be overriden.

tools/h5part_reader.py
adigitoleo: It doesn't support appending, so that would overwrite the result of `mineral.save()` from two lines above.

tools/h5part_reader.py
adigitoleo: Yes, the CPO data is stored on particles in a scalar array with dimension ~10000 (10 * n_grains), which is written out to 10000 single-element arrays in the h5part file. I think this is why the h5part files are so large.

Patol75 (approved):

PR #72, opened 2023-02-22
fix: Allow specifying  kwargs in  call

adigitoleo:
Fixing the API for `polefigures`


adigitoleo:
Not quite, that needs some more work.

PR #78, opened 2023-03-07
fix: Fix dislocation density equation

adigitoleo:
Calculation of dislocation density was incorrect: 

- the prefactor should be (1/crss[i]), since crss[i] = τ_s/τ_0 in equation 16, Fraters & Billen 2021 

- the exponent for the second term should be p/n rather than n/p (i.e. it should be stress_exponent/deformation_exponent) see also eq. 16, Fraters & Billen 2021

Note that the closest equation in the original paper, eq. 22 of Kaminski & Ribe 2001, is a bit confusing in that it uses the notation crss[i] = τ_0, where both Fraters and I have adopted crss[i] = τ* = τ/τ_0 and reserved τ_0 for the specific crss value of the softest slip system.

Note also that eq. 11 in Kaminski & Ribe 2004, which is supposed to be a reprinting of this same equation (except now τ_0 is something different), adds a spurrious division by the strain rate scale, as documented in the code.


adigitoleo:
There is also some renaming happening to make the values `stress_exponent` and `deformation_exponent` in the code match what the docs in the `__init__.py` file are saying.

adigitoleo:
Note that from what I can see the bug also affected the initial translation:
https://github.com/Patol75/PyDRex/blob/main/src/initial_implementation/PyDRex.py#L437

On that line the prefactor is `tau[index] instead of `(1 / tau[index])`

Patol75:
If I understand properly, nothing changes here: this is purely name swapping and term ordering. Still, can you remind me of where `crss` is defined in your code?

adigitoleo:
> where `crss` is defined

You mean in documentation? https://patol75.github.io/PyDRex/pydrex/minerals.html#L-4

Patol75:
I meant as a function of tau.

Patol75 (approved):

PR #81, opened 2023-03-14
More documentation fixes

adigitoleo:
- Make documentation of CRSS more precise
- Add missing enum docstrings
- Promote comments to docstrings in core tests
- Make print statements into logging calls

### DIFF
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -82,7 +82,6 @@ def get_crss(phase, fabric):
     raise ValueError("phase must be a valid `MineralPhase`")
 
 
-@nb.njit
 def get_primary_axis(fabric):
     """Get primary slip axis name for the given olivine `fabric`."""
     match fabric:
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -1,8 +1,9 @@
-"""> PyDRex: Lagrangian data structures and fabric/CRSS selection.
+"""> PyDRex: Lagrangian data structures and mineral fabric definitions.
 
-Acronyms:
+**Acronyms:**
 - CRSS = Critical Resolved Shear Stress,
-    i.e. components of stress acting on each slip system in the grain reference frame
+    i.e. threshold stress required to initiate slip on a slip system,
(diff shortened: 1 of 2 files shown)


src/pydrex/minerals.py
```diff
@@ -1,8 +1,9 @@
-"""> PyDRex: Lagrangian data structures and fabric/CRSS selection.
+"""> PyDRex: Lagrangian data structures and mineral fabric definitions.
 
-Acronyms:
+**Acronyms:**
 - CRSS = Critical Resolved Shear Stress,
-    i.e. components of stress acting on each slip system in the grain reference frame
+    i.e. threshold stress required to initiate slip on a slip system,
+    normalised to the stress required to initiate slip on the softes slip system
```
Patol75, line 6: softest

Patol75 (changes requested):
Found one typo, the rest looks good!

Patol75 (approved):

Patol75:
It seems like we have a test failure. Any merit in replacing some of these `assert` statements by [numpy.testing](https://numpy.org/doc/stable/reference/routines.testing.html) functions to get more information on failures? `np.testing.assert_allclose()` should be a good replacement for `assert np.isclose()`.

adigitoleo:
> test failure

I've edited the tolerance value and will submit that in the next test fixes. Apologies, I re-ran the job before reading your comment. The failure was once again due to one of the output diagnostics being slightly different and slightly past the tolerance.

There are two sources of randomness in the outputs:
- Random orientations in initial condition are not deterministic. Maybe I should seed the RNG for the tests.
- The diagnostics (M-index, angle) rely on re-sampled orientations, which once again uses an RNG. Maybe I should also seed that one for now, ultimately I would like for it to converge a bit better and am testing now with different number of grains, timesteps, etc.

> replacing some of these `assert` statements by [numpy.testing](https://numpy.org/doc/stable/reference/routines.testing.html) functions

I guess you mean using assert_allclose against the entire `misorient_angles`/`misorient_indices` arrays, instead of just assert statements against specific values? Currently that would make the test even more flaky, but it's probably a better approach once I seed the RNGs.

PR #82, opened 2023-03-21
Finalise SCSV format

adigitoleo:
Full SCSV I/O and tests. These files will not only be used for third party datasets but also to save pathlines from standalone runs, to complement the NPZ files which only store CPO information.

### DIFF
--- a/data/specs/spec.scsv
+++ b/data/specs/spec.scsv
@@ -3,13 +3,13 @@
 # The SCSV format is a plain text format intended for small data files.
 # A YAML header with a data schema and optional comments is added to a standard CSV file.
 # This idea was inspired by CSVY (http://csvy.org/) but does not aim to be compatible with that format.
-# A single SCSV file can store exactly one dataset.
+# A single SCSV file can store exactly one, flat dataset.
 # The file encoding must be UTF-8.
 
 # The YAML header must obey the following structure:
-schema:
+schema:  # The delimiter and missing string definitions must be enclosed in quotes, e.g. ','.
   delimiter: ','  # A single character delimiter.
-  missing: '-'  # An arbitrary-length string denoting missing data.
+  missing: '-'  # An arbitrary-length string denoting missing data (must not contain the delimiter).
   fields:  # Inline comments are supported in the YAML section.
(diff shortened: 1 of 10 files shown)


data/specs/spec.scsv
```diff
@@ -0,0 +1,37 @@
...
+  delimiter: ','  # A single character delimiter.
+  missing: '-'  # An arbitrary-length string denoting missing data.
+  fields:  # Inline comments are supported in the YAML section.
+    - name: first_column  # Column names must be valid Python identifiers as per `str.isidentifier()`.
+      type: string  # Valid types are 'string', 'integer', 'float', 'boolean' or 'complex' (number).
+      fill: "MISSING"  # Fill value for missing data, replacement for the 'missing' string.
+      unit: percent  # Arbitrary, optional string describing the data units, currently not parsed.
+    - name: second_column  # Type and fill value are also optional, defaults are 'string' and "".
+    - name: third_column
+      type: integer  # Type of fill value must match the specified type. Use 'NaN' for floats and complex numbers.
```
Patol75, line 20: `Use 'NaN' for floats and complex numbers.` What is the reasoning here?

Patol75 (dismissed):

data/specs/spec.scsv
adigitoleo: NaN is a standard(ised) floating point value, so array of floats can have NaN (Python np.nan) without sacrificing type homogeneity. For complex numbers it's a bit more of a stretch, but `complex(np.nan)` is valid Python and gives a value with np.nan as the real component and float(0) for the imaginary component. I think that still conveys the intended meaning of "this value is not valid/missing data".

However, it is possible to use other values to fill missing data, for example `0.0` for floats or `0`/`0.0`/`0j` for complex, which all become `0j` in the parsed data.

Overall the format is intended to be easy to parse while staying simple for human reading. I'll edit that spec comment to say that, in addition to fill values of the given type, you can also use `"NaN"` which will be interpreted properly.

adigitoleo:
Woops, forgot to commit the test file, not ready for merge until I get home and add that.

Patol75 (approved):

PR #84, opened 2023-03-23
Fixes to test configuration and CPO integration

adigitoleo:
- seed for RNG
- fix `--outdir` fixture scope
- attempt to fix the pathline and CPO integrators
- save pathlines to output files for better debugging

What is left to do:
- [x] : Replace the asserts in the 2d corner flow test, the values changed because of the RNG seed but should be reproducible now


src/pydrex/minerals.py
```diff
@@ -336,19 +336,14 @@ def apply_gbs(orientations, fractions, config):
                 )
             ),
             time_end,
-            # first_step=max_step / 4,  # TODO: Move divisor to config?
-            max_step=max_step,
-            # atol=np.inf,  # FIXME: Different solver or smart way to set tolerance.
+            # first_step=max_step / 4,
+            # max_step=max_step,
+            rtol=1e-6,
```
Patol75, line 341: You may want to set `atol` to 0. The reasoning here is that `atol` can bypass `rtol` if the values evaluated in the integration process are smaller than `atol`.

src/pydrex/pathlines.py
```diff
@@ -64,10 +64,11 @@ def _max_strain(
...
-        method="RK45",  # TODO: Compare to LSODA?
-        # first_step=1e10,
+        method="LSODA",
+        first_step=1e10,
         max_step=np.inf,
         # max_step=1e6,
+        t_eval=None,
         events=[_max_strain],
         args=(interp_velocity, interp_velocity_gradient, min_coords, max_coords),
         dense_output=True,
```
Patol75, line 74: Same idea here with the tolerances?

tests/test_corner_flow_2d.py
```diff
@@ -211,29 +213,29 @@ def _get_velocity_gradient(point):
...
+                #         assert misorient_angles[-1] < 20
+                #     case -0.54:
+                #         assert misorient_indices[halfway] > 0.575
+                #         assert misorient_angles[halfway] > 22
+                #         assert misorient_indices[-1] > 0.7
+                #         assert misorient_angles[-1] < 22
+                #     case -0.78:
+                #         # TODO: Results from this pathline are a bit weird.
+                #         # No tests for now...
+                #         pass
```
Patol75, line 238: Why did you comment all the `assert` statements?

Patol75 (commented):
Looks good, so long as the issues with the corner-flow test are resolved. A few comments on my side.

tests/test_corner_flow_2d.py
adigitoleo: Becuase of the RNG seed, the `angles`, `indices`, etc. should now be reproducible, so I'm going to replace those somewhat arbitrary assserts with some `np.testing.assert_allclose` or statements instead.

src/pydrex/minerals.py
adigitoleo: Have set `atol` to 1e-99, I found that 0 caused occasional crashes depending on n_grains/timesteps.

src/pydrex/pathlines.py
adigitoleo: A few lines below, atol and rtol are set. Maybe those values can be changed, or that could be in another PR after I check LSODA vs RK45 for both pathline and CPO integrators.

adigitoleo:
The other test (simple shear) doesn't like CPO integrator having `RK45`, `atol=1e-99`, `rtol=1e-6`, which causes small timesteps again and eventually a crash. Possibly solved with `first_step` or by using `LSODA`, will continue to investigate.

adigitoleo:
`atol` and `rtol` set to defaults as discussed before Easter, can also be changed by the user using keyword arguments to `update_orientations`. This may be a bit on the slow side for now but better to be accurate at the moment and optimise later if possible.

Patol75 (approved):

Issue #87, opened 2023-03-28
add pole figure contouring


adigitoleo:
We need contouring of pole figures, like [here](https://github.com/MFraters/cpo_analyzer).

No quick and easy way to do this, need manual counting stations on the circle and the color corresponds to the number of counts "near" the counting station/point compared to the number expected for a uniform distribution.

Then apply smoothing, see [here](https://github.com/joferkington/mplstereonet/) for some examples using linear/exponential smoothing etc.

adigitoleo:
Rudimentary contouring introduced in #93, very sensitive to kernel smoothing parameter.

adigitoleo:
Proper contouring depends on ODF, see #113.

adigitoleo:
https://github.com/vollmerf/spherecontour

Patol75:
https://vollmerf.github.io/orient/

PR #93, opened 2023-05-03
Pole figures

adigitoleo:
Adds pole figure contouring using a counting method. Kernel functions are implemented according to [Vollmer 1995](https://doi.org/10.1016/0098-3004(94)00058-3), which are also used in [mplstereonet](https://github.com/joferkington/mplstereonet/), although here the counting grid and default smoothing parameters are different, and I've implemented a simpler Axes subclass that doesn't override the default plotting methods. Instead, it provides the additional method `polefigure` which handles all pole figure plotting. This way we are not affected by the awkward `matplotlib.transforms` API and don't have any trouble with their regressions.

There are a few extra things in here that I didn't end up using but may be useful later, like the Shirley square to disk equal-area projection.

In general, it still looks slightly different to MTEX and others but is probably good enough for qualitative comparison. Later it would be ideal to add some kind of 3D likelihood estimation using [von Mises UARS](https://doi.org/10.1198/jasa.2009.ap08741) or [Bingham](https://github.com/sebastianriedel/bingham) distributions, and then plot the fitted distribution instead of doing this manual counting + ad hoc smoothing. Currently the contoured pole figures are highly sensitive to the smoothing parameter. Apparently in MTEX, they use a spherical harmonic representation instead, and even there they had to add [a sophisticated smoothing parameter optimisation](https://mtex-toolbox.github.io/DensityEstimation.html). [The code that Menno used](https://github.com/MFraters/cpo_analyzer) does some simple counting like we do here, but I think he uses simple Gaussian kernels not the Kamb kernels from Vollmer 1995 (which may be less sensitive to sample size i.e. number of grains).

#87

Some plots, to show how it looks:
polefigures polefigures_raw

pf_contoured


tests/test_geometry.py
```diff
@@ -0,0 +1,39 @@
...
+    np.testing.assert_allclose(ref_data["zvals"], zvals, atol=1e-16, rtol=0)
+
+
+def test_lambert_equal_area(rng):
+    """Test Lambert equal area projection."""
+    a = [x / 5.0 for x in range(-5, 6)]
+    x = [[x] * len(a) for x in a]
+    y = [a for _ in a]
+    x_flat = [j for i in x for j in i]
+    y_flat = [j for i in y for j in i]
```
Patol75, line 29: Surely this can be achieved with `numpy.tile` and `numpy.repeat`, given that you import `numpy`.

src/pydrex/geometry.py
```diff
@@ -0,0 +1,170 @@
...
+    zvals = np.atleast_1d(zvals).astype(float)
+    # One reference for the equation is Mardia & Jupp 2009 (Directional Statistics),
+    # where it appears as eq. 9.1.1 in spherical coordinates,
+    #   [sinθcosφ, sinθsinφ, cosθ].
+
+    # First we move all points into the upper hemisphere.
+    # See e.g. page 186 of Snyder 1987 (Map Projections— A Working Manual).
+    # This is done by taking θ' = π - θ where θ is the colatitude (inclination).
+    # Equivalently, in Cartesian coords we just take abs of the z values.
+    zvals = abs(zvals)
```
Patol75, line 92: Does the 0 value of `zvals` initially correspond to the North Pole and all other values are negative?

src/pydrex/geometry.py
```diff
@@ -0,0 +1,170 @@
...
+    directions[:, 2] /= directions_norm
+
+    # Rotate into the chosen reference frame.
+    zvals = directions[:, axes_map[upward_axes]]
+    yvals = directions[:, axes_map[ref_axes[1]]]
+    xvals = directions[:, axes_map[ref_axes[0]]]
+    return xvals, yvals, zvals
+
+
+def lambert_equal_area(xvals, yvals, zvals):
```
Patol75, line 70: Is there any similar function in `pyproj`?

src/pydrex/stats.py
```diff
@@ -176,3 +177,152 @@ def _max_misorientation(system):
...
+    **kwargs,
+):
+    """Estimate point density of orientation data on the unit sphere.
+
+    Estimates the density of orientations on the unit sphere by counting the input data
+    that falls within small areas around a uniform grid of spherical counting locations.
+    The input data is expected in cartesian coordinates, and the contouring is performed
+    using kernel functions defined in [Vollmer 1995](https://doi.org/10.1016/0098-3004(94)00058-3).
+    The following optional parameters control the contouring method:
+    - `gridsteps` (int) — the number of steps, i.e. resolution of the spherical counting grid
```
Patol75, line 199: Perhaps say "the number of points on the spherical counting grid"?

Patol75 (changes requested):
Super cool stuff, well done! Some comments and questions for you to address.

src/pydrex/geometry.py
adigitoleo: There is the `laea` projection EPSG code 9820, however the pyproj functions take spherical coord inputs and don't perform the hemisphere flipping. I'm not sure it's worth it to pull that dependency for one function when I would have to do almost the same amount of preprocessing to give it the data in the right conventions. I'm also not sure if they have plain spherical projections rather than using Earth ellipsoids (although it would be surprising if they didn't).

src/pydrex/geometry.py
adigitoleo: z value of +1 is the "north pole" or the upward pointing axis. 0 is at the equator. -1 at the "sout pole", so all negative values are flipped to make the 3d vectors all point to the upper hemisphere, if that makes sense.

src/pydrex/geometry.py
```diff
@@ -0,0 +1,170 @@
...
+    directions = np.tensordot(orientations.transpose([0, 2, 1]), hkl, axes=(2, 0))
+    directions_norm = la.norm(directions, axis=1)
+    directions[:, 0] /= directions_norm
+    directions[:, 1] /= directions_norm
+    directions[:, 2] /= directions_norm
+
+    # Rotate into the chosen reference frame.
+    zvals = directions[:, axes_map[upward_axes]]
+    yvals = directions[:, axes_map[ref_axes[1]]]
+    xvals = directions[:, axes_map[ref_axes[0]]]
```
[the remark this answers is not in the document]
adigitoleo: It's just a right-handed coord. frame, so as per the docstring it expects `ref_axes` to be given in that particular order (second and third of the right-handed axes).

tests/test_geometry.py
adigitoleo: Actually mgrid is fine. I think I had it like this because it was initially in a docstring example but I just copied it across. Fixed now.

Patol75 (approved):

PR #98, opened 2023-06-19
Fixes to strain energy algo and CPO solver

adigitoleo:
Main important changes are the ones in:

- src/pydrex/core.py: strain energy fix
- src/pydrex/minerals.py: adds first_step to CPO solver and refactors so that NaNs do not appear in `y` in eval_rhs (scipy doesn't like having extra args in the eval_rhs function, although I couldn't come up with a specific reproducer without using our particular PDEs)

The rest updates the testing/build boilerplate to make it all work. Probably best to squash.


Patol75 (approved):
All looks good! That fix in `core.py` will avoid us a few headaches when switching to other olivine fabrics!

Issue #104, opened 2023-07-03
Fix Numba typing errors for multiphase cases


adigitoleo:
Without https://github.com/Patol75/PyDRex/commit/3d6a8729661f7dcf4afe8679d650bd6c3080824d, Numba throws typing errors because the JIT compiler struggles with the Enums (can't distinguish A-type olivine (0) from "A-type" (0) (the only type of) enstatite?) This needs a regression test after the fix is merged.

adigitoleo:
Fixed in #107

adigitoleo:
No need for regression test anymore because the Enum was refactored and enstatite is now a different number (5) compared to olivine A-type (0)

PR #107, opened 2023-07-09
Add core OPX tests

adigitoleo:
Adds some enstatite tests to catch a Numba typing issue when multiphase minerals were used, and refactors some symbols to avoid the crash.

This refactoring avoids the need for explicit numba type signatures which led to more numba errors for other tests.

### DIFF
--- a/src/pydrex/core.py
+++ b/src/pydrex/core.py
@@ -169,17 +169,12 @@ def _get_deformation_rate(phase, orientation, slip_rates):
     deformation_rate = np.empty((3, 3))
     for i in range(3):
         for j in range(3):
-            if phase == MineralPhase.olivine:
-                deformation_rate[i, j] = 2 * (
-                    slip_rates[0] * orientation[0, i] * orientation[1, j]
-                    + slip_rates[1] * orientation[0, i] * orientation[2, j]
-                    + slip_rates[2] * orientation[2, i] * orientation[1, j]
-                    + slip_rates[3] * orientation[2, i] * orientation[0, j]
-                )
-            elif phase == MineralPhase.enstatite:
-                deformation_rate[i, j] = 2 * orientation[2, i] * orientation[0, j]
-            else:
-                assert False  # Should never happen.
+            deformation_rate[i, j] = 2 * (
+                slip_rates[0] * orientation[0, i] * orientation[1, j]
+                + slip_rates[1] * orientation[0, i] * orientation[2, j]
+                + slip_rates[2] * orientation[2, i] * orientation[1, j]
+                + slip_rates[3] * orientation[2, i] * orientation[0, j]
+            )
     return deformation_rate
 
 
@@ -210,6 +205,10 @@ def _get_slip_rate_softest(deformation_rate, velocity_gradient):
             enumerator += 2 * deformation_rate[j, L] * velocity_gradient[j, L]
             denominator += 2 * deformation_rate[j, L] ** 2
 
+    # Tiny denominator means that relevant deformation_rate entries are zero.
+    # No deformation rate means no slip rate.
+    if -1e-15 < denominator < 1e-15:
+        return 0.0
     return enumerator / denominator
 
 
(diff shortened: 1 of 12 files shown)


adigitoleo:
Closes #104

src/pydrex/core.py
```diff
@@ -1,19 +1,92 @@
-r"""> PyDRex: Core D-ReX functions.
+r"""> PyDRex: Core D-ReX functions and enums.
```
Patol75, line 1: Not that important, but I think it is D-Rex, not D-ReX

src/pydrex/core.py
```diff
@@ -1,19 +1,92 @@
...
 
-CPO can be calculated along a flow pathline, using interpolated velocity values,
-or for a Lagrangian particle experiencing a given velocity gradient.
-
-.. warning::
-    It is recommended to use the `Mineral` class from `pydrex.minerals`
-    instead of using these routines directly, which do not circumvent all edge cases.
-    For example, the pathological case with a flow field with zero vorticity will crash.
+The function `derivatives` implements the core DRex solver, which computes the
```
Patol75, line 3: D-Rex

src/pydrex/core.py
```diff
@@ -381,3 +372,88 @@ def _get_strain_energy_enstatite(
...
+            slip_invariants,
+            slip_indices,
+            crss,
+            deformation_exponent,
+        )
+    elif phase == MineralPhase.enstatite:
+        slip_indices = np.argsort(1 / crss)
+        slip_rates = np.repeat(np.nan, 4)
+    else:
+        assert False  # Should never happen.
```
Patol75, line 428: Just a little preference here, but perhaps `raise SystemExit` instead of `assert False`? It seems more explicit to me.

tests/test_core.py
```diff
@@ -7,10 +7,92 @@
...
+            for θ in np.mgrid[0 : 2 * np.pi : 360j]:
+                _log.debug("θ (°): %s", np.rad2deg(θ))
+                orientations_diff, fractions_diff = _core.derivatives(
+                    phase=_core.MineralPhase.enstatite,
+                    fabric=_core.MineralFabric.enstatite_AB,
+                    n_grains=1,
+                    orientations=Rotation.from_rotvec([[0, 0, θ]]).as_matrix(),
+                    fractions=np.array([1.0]),
+                    strain_rate=np.array([[0, 1, 0], [1, 0, 0], [0, 0, 0]]),
+                    velocity_gradient=np.array([[0, 2, 0], [2, 0, 0], [0, 0, 0]]),
```
Patol75, line 77: Is the velocity gradient correct here?

Patol75 (changes requested):
Just a few items to check and should be good to go!

src/pydrex/core.py
adigitoleo: OK good idea, I will use `SystemExit(1)` with code 1 for failure for these "programmer error" errors (I'm not throwing a normal Exception here because it's not an user error).

tests/test_core.py
adigitoleo: good catch

src/pydrex/core.py
```diff
@@ -364,22 +377,17 @@ def _get_strain_energy_enstatite(
     - `nucleation_efficiency` (float) — parameter controlling grain nucleation
 
     """
-    weight_factor = slip_rate_softest / crss[slip_indices[-1]] ** stress_exponent
+    # weight_factor = slip_rate_softest / crss[slip_indices[-1]] ** stress_exponent
     dislocation_density = (1 / crss[slip_indices[-1]]) ** (
         deformation_exponent - stress_exponent
-    ) * np.abs(weight_factor) ** (stress_exponent / deformation_exponent)
+    ) * np.abs(slip_rate_softest) ** (stress_exponent / deformation_exponent)
```
Patol75, line 380: What is happening here?

Patol75 (commented):
Looking good; just have one question for you.

src/pydrex/core.py
adigitoleo: Ah, I had moved the division by that factor up to an earlier line in preparation for unifying the strain energy functions (since that is the only difference between the one for enstatite and olivine). But forgot to actually merge the functions. That's done in the new commit and this calculation is moved to line 413.

PR #116, opened 2023-07-20
Update 2D simple shear CI tests

### LINKED ISSUE #110: A-type olivine in 2D simple shear
Tracking issue for 2D simple shear benchmark status.

For 2D simple shear, we should be able to:
- [x] reproduce Kaminski & Ribe, 2001, fig. 5 results (more or less, although not always within 1 standard devation of our ensemble...)
- [x] observe the effect of M* to increase the speed of CPO to shear alignment
- [x] ~~reproduce the Kaminski & Ribe, 2004, fig. 4 results~~ Can't reproduce, giving up because not enough info on timestep resolution etc.
- [x] observe the effect of grain boundary sliding to impede unphysically high texture strength
- [x] observe a secondary orientation peak in the pole figures (at about 45° from the main one in the [100] pole figures)
- [x] observe a dependence on initial texture condition (e.g. Boneh 2015)
- [x] observe a match at large strains (>t*=0.5) in FSE axis and CPO axis evolution when M*=0

Diagnostic figures:

Pole figures for M*=200, no sliding:

Pole figures for X=0.4, second peak at 45deg stays present at high strain in agreement with Fig. 1 from Kaminski 2004:

adigitoleo:
Closes #110.

Uses symmetry diagnostic instead of expensive strength diagnostics which still require further clarification.

Need to convince ourselves that the remaining differences from Kaminski 2001 and 2004 are justified:
- For the GBM test, max. difference is about 5.7 degrees, we don't reach 0 degrees at the 200% strain cutoff
- For the GBS test, the parameter is mostly influencing point-symmetry but only slightly slowing down rotation of the mean a-axis, not by 10+ degrees after 200% strain. Our a-axis also rotates immediately without the initial stagnation (there is no theoretical grounds for the initial stagnation I believe).


adigitoleo:
Quirky IndexError reproduces on cosgrove (ubuntu, Python 3.11.3) but not for me (Arch and Void, 3.11.4).

src/pydrex/minerals.py
```diff
@@ -358,8 +362,7 @@ def apply_gbs(orientations, fractions, config):
                 len(np.nonzero(mask)) / len(fractions),
             )
             # No rotation: carry over previous orientations.
-            # TODO: Should we really be resetting to initial orientations here?
-            orientations[mask, :, :] = self.orientations[0][mask, :, :]
+            orientations[mask, :, :] = self.orientations[-1][mask, :, :]
```
Patol75, line 365: This changes from FORTRAN D-Rex, right?

src/pydrex/utils.py
```diff
@@ -0,0 +1,13 @@
+"""> PyDRex: Miscellaneous utility methods."""
+import numpy as np
+
+
+def skip_nans(a):
+    """Skip NaN values in array."""
```
Patol75, line 6: Skip or remove?

Patol75 (commented):
Getting there! Let me have a look at the CI failure.

src/pydrex/minerals.py
adigitoleo: Not from what I can tell, at least from the version on the website (not sure about Chris' or Navid's version). In the original one, the sliding parameter (chi) only appears 4 times in the code, on the lines 66 (declaration), 388 (read value from input file), 1385 and 1460 (sliding for olivine and enstatite, respectively). In the last two cases, the context is identical (just swapping dotacs -> dotacs_ens and rt -> rt_ent).

```
!!! grain boundary sliding for small grains
   IF (odfi(i) .LT. chi/REAL(size)) THEN
      dotacs(i,:,:) = 0d0
      rt(i) = 0d0
   END IF
```

In other words, they set the differential of the orientation to zero, not the orientation itself (and also the dislocation density `rt`). In our code that would be the orientations_diff, not orientations. Setting that to zero is equivalent to carrying over the previous orientation as I have done here.

src/pydrex/utils.py
adigitoleo: Well, Julia has `skipmissing` which removes `missing` values :joy: I'll rename it though.

Patol75:
Current CI reports two minimal `rtol` violations and two memory allocation errors. For the latter, they are likely related to the number of grains increasing from 3375 to 4394. Either decrease it or use `float32` for that specific test; what do you think?

adigitoleo:
> Current CI reports two minimal `rtol` violations and two memory allocation errors. For the latter, they are likely related to the number of grains increasing from 3375 to 4394. Either decrease it or use `float32` for that specific test; what do you think?

It's one of each, the pytest "capture" output makes errors look duplicated. The number of grains matches the value in the GBS paper (Kaminski & Ribe, 2004, section 4.1 second line) but since there are bigger issues at the moment with matching the GBS results maybe decreasing is fine.

src/pydrex/minerals.py
Patol75: Fair. I cannot quite find the updated FORTRAN file, but I am pretty sure they changed that bit. Have you checked what is implemented in ASPECT?

adigitoleo: It makes no sense to reset to index 0 at all, what if the initial condition is not isotropic?

Aspect implementation checks the greater than condition instead, and if the check fails then they simply set the strain energy to zero and leave the orientations unchanged which is like using index [-1] here not [0]:

https://github.com/geodynamics/aspect/blob/57d11548c089c3d44b40a84714e8c8f0e1938b3e/source/particle/property/crystal_preferred_orientation.cc#L781-L795

adigitoleo:
> Either decrease it or use `float32` for that specific test; what do you think?

Because the tests are now converted to ensemble tests that take a long time, I have disabled them by default and they must be enabled by passing the `--runslow` flag to pytest, so for now I'm leaving the number of grains as is (running the ensemble has higher memory requirements anyway).

Patol75 (approved):

PR #121, opened 2023-07-25
Multiprocessed ensemble 2D shear test (for `--runslow`)

adigitoleo:
Adds multiprocessing to the 2D ensemble shear tests and uses a smaller set of seeds by default, so that `--runslow` is feasible on a moderately powerful personal computer (16 cores, <16GB RAM, ~30 min). About 10h CPU time:

```
INFO [01:29] pydrex: elapsed CPU time: 10:00:24
```

After it runs for me I'll uncomment and fix the asserts. Looks similar to the 1000 seeds case but less resolved.

- [x] Add asserts
- [x] Fix test_config (CI failure)
- [x] ~~Fix FSE fluctuations (unfortunately still happens, not every run)~~ Can't reproduce reliably with these tests, will perhaps use another test to track that down later
- [x] See if it changes with `_USE_ORIGINAL_DREX=True` (not appreciably)


adigitoleo:
Closes #110, this time I won't reopen it :)

src/pydrex/io.py
```diff
@@ -168,73 +184,162 @@ def save_scsv(file, schema, data, **kwargs):
             writer.writerow(row)
 
 
-def parse_params(file):
-    """Parse an INI file containing PyDRex parameters."""
-    config = configparser.ConfigParser()
-    configpath = resolve_path(file)
-    config.read(configpath)
+def parse_config(path):
+    """Parse an TOML file containing PyDRex configuration."""
```
Patol75, line 188: `a TOML`

src/pydrex/velocity_gradients.py
```diff
@@ -0,0 +1,42 @@
+"""> PyDRex: Steady-state solutions of velocity gradients for various flows."""
+import functools as ft
+
+import numba as nb
+import numpy as np
+
+
+@nb.njit(fastmath=True)
+def _simple_shear_2d(x, direction=1, deformation_plane=0, strain_rate=1):
```
Patol75, line 9: variable `x` is not used; is that on purpose?

tests/test_simple_shear_2d.py
```diff
@@ -377,81 +458,246 @@ def get_velocity_gradient(x):
...
+                [
+                    np.median(m.fractions[halfway])
+                    - np.median(minerals[i].fractions[halfway])
+                    for i, m in enumerate(minerals[:-1], start=1)
+                ]
+            )
+            > 0
+        )
+
+    def test_boudary_sliding(self, seed, outdir):
```
Patol75, line 615: `boundary`

Patol75 (changes requested):
Found a couple of typos, but otherwise spot-on!

src/pydrex/velocity_gradients.py
adigitoleo: Yes, intended, for now only uniform shear flow. The 'x' is required just to catch the position arg that is passed to the function inside `update_orientations` (which I designed so that it can take position-dependent velocity gradients at some point).

Patol75 (approved):

PR #127, opened 2023-08-03
fix: Implement banded Jacobian for n_grains > 4632

### LINKED ISSUE #89: Settle on default pathline and CPO integrator options
It is tricky to find good default options for the scipy integrators in `Mineral.update_orientations` and `pathlines.get_pathline`. We need to settle on some defaults that should "work" (at least not crash) in common situations, and probably provide an interface to quickly/easily change those options.

Current status (checkboxes for when resolved):
- [x] ~~RNG seed of 13 causes GBS test with X=0.2 to hang at step 5/250~~ (nope, my mistake)
- [x] ~~RNG seed of 8845 causes hang at 347th matrix? (check if that persisted)~~ Can't reproduce with current atol/rtol/first_step settings, maybe solved
- [x] Can't handle certain numbers of grains ~~(or is it just memory?)~~ Too many ODE's for SciPy, solved by be0b15b

Pathline integrator:
- `RK45`, `atol=1e-99`, `rtol=1e-6`?

CPO integrator:
- `LSODA`, `atol=1e-99`, `rtol=1e-6`?

adigitoleo:
Closes #89

Uses a default bandwidth of 12000 (lband = uband = 6000), and exposes the options to the user via Mineral() and Mineral.update_orientations().

### DIFF
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -217,6 +217,8 @@ class Mineral:
     fractions: list = field(default_factory=list)
     orientations: list = field(default_factory=list)
     seed: int = None
+    lband: int = None
+    uband: int = None
 
     def __str__(self):
         # String output, used for str(self) and f"{self}", etc.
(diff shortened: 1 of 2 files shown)


adigitoleo:
Test fails because it runs out of memory of course (forgot about CI memory limit). Will create a new pytest mark for high-memory tests.

src/pydrex/minerals.py
```diff
@@ -254,6 +256,18 @@ def __post_init__(self):
             self.orientations_init = Rotation.random(
                 self.n_grains, random_state=self.seed
             ).as_matrix()
+        # For large numbers of grains, the number of ODE's is too large for SciPy.
+        # Instead, we need to solve the banded system of equations.
+        # By default, we use a bandwidth o f 12000 with lband = uband = 6000.
```
Patol75, line 261: `ODEs`
`exceeds what LSODA can handle.` instead of `is too large for SciPy.`
`Therefore, we specify the Jacobian matrix as banded.
`of`

Patol75 (changes requested):
Thanks for that. Just some wording in the comments. :)

src/pydrex/minerals.py
adigitoleo: OK, let me know if 3cdc4b0 is good.

PR #132, opened 2023-08-08
feat: Add legacy K&R2004 Runge Kutta solver

adigitoleo:
Adds a method that replicates the original RK solver, almost as a drop-in replacement for Mineral.update_orientations except a few extra arguments are required in the call signature.

### DIFF
--- a/src/pydrex/legacy.py
+++ b/src/pydrex/legacy.py
@@ -45,7 +45,7 @@ def update_orientations_Kaminski2001(
         else:
             assert False
 
-        for n in n_iter:
+        for n in range(n_iter):
             fsei = deformation_gradient
             odfi = mineral.fractions[-1]
             acsi = mineral.orientations[-1]
--- a/src/pydrex/legacy.py
+++ b/src/pydrex/legacy.py
@@ -11,17 +11,17 @@ def update_orientations_Kaminski2001(
     deformation_gradient,
     get_velocity_gradient,
     pathline,
-    faux_gridres,
+    gridstep,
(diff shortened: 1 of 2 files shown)


adigitoleo:
Don't merge yet, it needs a n_grains loop in there.

src/pydrex/legacy.py
```diff
@@ -0,0 +1,193 @@
...
+    get_velocity_gradient,
+    pathline,
+    faux_gridres,
+    velocity_magnitude,
+    n_steps,
+    **kwargs,
+):
+    """Implements the original RK45 approach (subroutine STRAIN).
+
+    Almost a drop-in replacement for `pdyrex.Mineral.update_orientations`,
```
Patol75, line 21: pydrex

src/pydrex/legacy.py
```diff
@@ -0,0 +1,193 @@
...
+    but requires a few additional arguments:
+    - `faux_gridres`: local 'grid resolution' or an approximation of the spatial
```
Patol75, line 23: What do you mean by `faux`?

Patol75 (commented):
Should be good to go :)

src/pydrex/legacy.py
adigitoleo: Changed to gridstep

Patol75 (dismissed):

Patol75:
We are getting some `IndexError` following the changes to the test file.

Patol75 (approved):

PR #133, opened 2023-08-08
feat: Add legacy eigh from the Fortran

adigitoleo:


adigitoleo:
Added some fixes to the legacy RK (so it matches the original more closely). Note that they don't update the velocity gradient at each internal RK iteration, it's just read once from the current "grid node". This is in contrast to how we have implemented `perform_step` et al. in `Mineral.update_orientations`, which evaluates `get_velocity_gradient` at each LSODA iteration.

Patol75 (approved):

PR #139, opened 2023-08-30
fix: Complete elastic decomposition and M-index and simplify API

adigitoleo:
Elastic decomposition now provides all symmetry component percentages.

M-index now takes into account the lattice system.
The maximum angle for random input still goes a bit over 120 degrees but
this might just be the tail of the distribution resulting from the
numerics? The mean seems to be slightly low for random input, around 80
instead of 90 degrees, but not too bad. Still, this method needs some
proper tests.

Most postprocessing functions should take a stack of orientations for a
series of texture snapshots, so NxMx3x3 orientations. In some cases
where the diagnostics are cheap to calculate this was not changed, since
it can be trivially achieved with list comprehension.


adigitoleo:
Some more notes:
- the components of the elastic decomposition look right when I check the Voigt matrices of the individual projections, however the percentages don't match Browaeys & Chevrot 2004 yet. The hexagonal % anisotrpy matches what comes out of the Fortran/original DREX code, which also doesn't match Browaeys & Chevrot 2004. I'll open an issue to flag this for later.
- the M-index should work now, well enough to attempt matching the values from Boneh et al 2015 for example. However, the equations from Grimmer to get theoretical misorientations for isotropic texture don't cover cubic crystals, which were apparently done in a previous work. Not important since olivine is orthorhombic but to be complete it could be extended to cover cubic sometime.

adigitoleo:
Ah, M-index tests need fixing, I don't think there's anywhere I can get raw CPO textures with associated M-index values (original paper has pole figures and M-indices but I'd have to see if Phil Skemer still has the data at some point). For now I'll just adjust the "expected" M-index values based on what I am able to get for say 1000 grains and then I'll bump those test down to a smaller number of grains to save CI time.

src/pydrex/diagnostics.py
```diff
@@ -195,40 +263,66 @@ def symmetry(orientations, axis="a"):
...
     The `bins` argument is passed to `numpy.histogram`.
     If left as `None`, 1° bins will be used as recommended by the reference paper.
-    The symmetry system can be specified using the `system` argument.
-    The default system is orthorhombic.
+    The `symmetry` argument specifies the lattice system which determines intrinsic
+    symmetry degeneracies and the maximum allowable misorientation angle.
+    See `_geo.LatticeSystem` for supported systems.
+
+    .. warning::
+        This method must be able to allocate an array of shape (N choose 2)x(M**2) for N
```
Patol75, line 307: What does `N choose 2` mean?

Patol75 (changes requested):
Massive work, well done!

src/pydrex/diagnostics.py
adigitoleo: I put the equation instead

adigitoleo:
Fails due to missing data files, will add them soon.

Patol75 (approved):

PR #140, opened 2023-09-04
Tools

adigitoleo:
- Add (parallel-capable) shell script to run Fortran 90 DRex ensemble
- Add CLI tool to print keys in a NPZ file


Patol75 (dismissed):

adigitoleo:
Had to change the Fortran DRex ensemble script because it didn't work as expected with multiple versions of DRex (different RNG seeds) reading from the same `comvar.mod`.

Patol75 (approved):

Issue #131, opened 2023-08-07
Sensitivity to number of grains


adigitoleo:
CPO results vary as a function of the number of grains, because they are essentially just samples in the orientation space, weighted by the `fractions`. Previous works have used anywhere from 500 to 5000 grains, with Fraters & Billen 2021 also testing 10000. Formal convergence tests are so far not available, except for some tangential discussions in the literature about the required number of samples for stable pole figure output.

Stability with respect to `n_grains` probably depends on the flow field to some extent, and will be affected by the value of the grain boundary sliding parameter.

Preview of convergence test in 2D simple shear for 100, 1000, 2000, 5000, 10000 grains. Shaded region is 1σ

![olivineA_ngrains_angles](https://github.com/Patol75/PyDRex/assets/34595875/6167569e-596c-4601-9e07-37e75ede8728)

Issue #117, opened 2023-07-21
Add nucleation efficiency sensitivity tests


adigitoleo:
While the original papers suggest $λ^{∗} = 5$ becase "the exact value between 3 and 50 does not significantly influence the LPO", this has not been observed by us here so far in practice. The parameter encodes not only the nucleation efficiency λ but in fact also the reference dislocation density which is poorly constrained. Our preliminary investigation suggests that changes in this
parameter on the order of 10 can change the CPO direction at 100% strain by the same order of magniture in degrees.

Because there is not enough experimental data to constrain this parameter any further we can only offer sensitivity tests so that calibration is made easier in future. The tests should describe how the model output changes for different values of this parameter. Probably simple shear considerations are sufficient for now.

adigitoleo:
basic case testing three values (3, 5, 10) added in #142

adigitoleo:
Done for now, GBM is more important.

PR #141, opened 2023-09-05
test: Rework 2D simple shear ensemble test

adigitoleo:
This reworks the ensemble test to be a parametrized set of ensemble tests for varying GBM, GBS and nucleation values. Plots show the effect of GBM for the chosen pair of GBS and nucleation, because GBM (M*) is the most important/least constrained of the three.

Ensemble tests don't compare to experiments nor the fortran, because these are targeted at testing the effects of the pydrex parameters specifically. I'll add a test back in vs the fortran and experiments soon.

Also here I only compute the angle diagnostic, because symmetry/strength diagnostics are expensive and don't add a lot here (I'll use separate more focused tests to compare to Boneh2015 or something).

5000 grains is good enough for steady simple shear, see #131 . I split it up using pytest parametrisation because that allows to run all cases simultaneously.

Nested `@pytest.mark.parametrize` works like `np.meshgrid`, giving all combinations. Using pytest `--collect-only` to list the expanded test list:
```shell
    <Function test_dvdx_ensemble[3-0]>
    <Function test_dvdx_ensemble[3-0.2]>
    <Function test_dvdx_ensemble[3-0.4]>
    <Function test_dvdx_ensemble[5-0]>
    <Function test_dvdx_ensemble[5-0.2]>
    <Function test_dvdx_ensemble[5-0.4]>
    <Function test_dvdx_ensemble[10-0]>
    <Function test_dvdx_ensemble[10-0.2]>
    <Function test_dvdx_ensemble[10-0.4]>
```

So each one can be selected individually in the PBS script with `-k="ensemble[3-0]"` etc.

### DIFF
--- a/tests/test_simple_shear_2d.py
+++ b/tests/test_simple_shear_2d.py
@@ -328,7 +328,6 @@ class TestOlivineA:
                     get_velocity_gradient,
                     shear_direction,
                     seed=seed,
-                    log_param="gbm_mobility",
                     return_fse=True,
                     ncpus=ncpus,
                 )
@@ -429,7 +428,6 @@ class TestOlivineA:
                     get_velocity_gradient,
                     shear_direction,
                     seed=seed,
-                    log_param="gbs_threshold",
                     return_fse=True,
                     ncpus=ncpus,
                 )
@@ -516,6 +514,5 @@ class TestOlivineA:
                     get_velocity_gradient,
                     shear_direction,
                     seed=seed,
-                    log_param="number_of_grains",
                     return_fse=True,
                 )
--- a/tests/test_simple_shear_2d.py
+++ b/tests/test_simple_shear_2d.py
@@ -14,6 +14,7 @@ from pydrex import diagnostics as _diagnostics
 from pydrex import io as _io
 from pydrex import logger as _log
 from pydrex import minerals as _minerals
+from pydrex import stats as _stats
 from pydrex import utils as _utils
 from pydrex import velocity_gradients as _dv
 from pydrex import visualisation as _vis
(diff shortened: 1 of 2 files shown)


adigitoleo:
#117

tests/test_simple_shear_2d.py
```diff
@@ -40,37 +41,32 @@ def run(
...
         shear_direction,
         seed=None,
-        log_param=None,
-        use_bingham_average=False,
-        return_fse=True,
-        ncpus=1,
+        return_fse=None,
     ):
-        """Reusable logic for 2D olivine simple shear tests.
+        """Reusable logic for 2D olivine A simple shear tests.
```
Patol75, line 46: What is going on here?

tests/test_simple_shear_2d.py
```diff
@@ -40,37 +41,32 @@ def run(
...
 
-        Always returns a tuple with 4 elements
-        (mineral, mean_angles, texture_symmetry, θ_fse),
-        but if `return_fse` is None then the last tuple element is also None.
+        Returns a tuple with the mineral and the FSE angle (or `` if `return_fse` is
```
Patol75, line 48: What do you mean by returning ``?

Patol75 (changes requested):
Looking good, just two small comments on my side.

tests/test_simple_shear_2d.py
adigitoleo: A-type Ol, typo

tests/test_simple_shear_2d.py
adigitoleo: Missing a `None`, typo

Patol75 (approved):

PR #142, opened 2023-09-07
Rework 2D shear GBM test

adigitoleo:
The 2D shear GBM test is now an ensemble test that compares ensemble
results from pypdrex with ensemble results from the Fortran for the same
parameters. The default number of grains in pydrex was raised again from
2500 to 3500 to be closer to the number originally used in DRex.

Using the 41 runs (`seeds_nearX45`) is enough to pass the checks, and
the test completes in about 10:30 CPU hours using a bit less than 15G
RAM. This can't be run in CI but can be run in less than an hour on
moderately powerful personal PCs (use the custom `--ncpus` arg to pydrex
to run in parallel).


tests/test_simple_shear_2d.py
```diff
@@ -321,98 +337,148 @@ def test_boundary_mobility(self, seed, ncpus, outdir):
...
-                [θ[-1] - angles[i][-1] for i, θ in enumerate(angles[:-1], start=1)]
-            )
-            > 0
+
+        # Check that results match Fortran output past 40% strain.
+        nt.assert_allclose(
+            result_angles[0][i_strain_40p:],
+            M0_drexF90[i_strain_40p:],
+            atol=0.1,
+            rtol=1,
```
Patol75, line 445: Are you sure you want `rtol=1` here?

Patol75 (changes requested):
Looks reasonable to me. Just one `rtol` that seems wrong.

tests/test_simple_shear_2d.py
adigitoleo: Thanks, fixed.

Patol75 (approved):

PR #144, opened 2023-09-21
fix: Fix default number of CPUs for parallel M-index calculation

adigitoleo:
Missing a call to `len` here.


Patol75 (approved):
Good catch, I missed it when I reviewed it!

PR #146, opened 2023-09-27
Core rotation rate and grain growth tests

adigitoleo:
Adds a couple of core tests for rotation rate and grain growth with initial grain orientations distributed on a circle (either in the plane of deformation or in the shear plane). The output figures offer some more clarity on what is happening with recrystallisation. For the first of these tests it was possible to derive analytical results for comparison. Unlike the ensemble 2D tests, these are fast enough to run in CI (a few seconds) because only the initial instantaneous rotation rate and grain growth values are checked.

This is the figure for the in-plane test. Both subparallel and suborthogonal grains want to grow in size, as well as a sharp peak at +/- 45 degrees from these directions. The rest of the grains have negative growth values which means they will shrink. The bottom figure shows rotation due to slip. Grains orthogonal to the shear plane rotate fastest, aligned grains do not rotate.

recrystallisation_2D_dvdx_circle_inplane

### DIFF
--- a/src/pydrex/utils.py
+++ b/src/pydrex/utils.py
@@ -68,6 +68,14 @@ def quat_product(q1, q2):
     ]
 
 
+def redraw_legend(ax):
+    """Redraw legend on matplotlib axis with new labels since last `ax.legend()`."""
+    legend = ax.get_legend()
+    if legend is not None:
+        legend.remove()
+    ax.legend()
+
+
 def __run_doctests():
     import doctest
 
--- a/src/pydrex/visualisation.py
+++ b/src/pydrex/visualisation.py
@@ -166,7 +166,7 @@ def alignment(ax, strains, angles, markers, labels, err=None, θ_max=90, θ_fse=
 
     if ax is None:
         fig = plt.figure(dpi=300)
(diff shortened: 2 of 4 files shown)


adigitoleo:
Not ready for merge, found one small issue in the second test.

adigitoleo:
OK now the second one is correct, had to use `from_euler` instead of `from_rotvec` because i need to rotate the crystal a-axes (X-axes) into the global YZ plane via composite rotation (rotate around Z by 90 to get into the YZ plane, then rotate by theta in that plane for the initial condition). Either way in this case the rotation rates and the grain growth is more complicated because multiple slip systems are active. Instead of comparing to analytics (if it is even available here) I just chec that the dominant slip system changes depending on the angular distance from the shear direction.

Grain growth in this case only occurs for subparallel orientations, all the rest will shrink. This explains how grains with oblique orientations in the shear plane disappear quickly despite not being able to slip on (010)[100] (i.e. why the left traces in the last row of Fig. 5 Fraters & Billen quickly go to zero compared to those on the right column).

recrystallisation_2D_dvdx_circle_shearplane

Patol75 (approved):

PR #148, opened 2023-09-28
test: Add GBM sanity check CI tests and mark cornerflow as slow

adigitoleo:
Also removed the redundant GBS ensemble (GBS example is covered by the big `test_dvdx_ensemble` example). Removed basic n_grains test which didn't have asserts, to be replaced with a more complete convergence test.


adigitoleo:
5 min CI is much more manageable than 2h :)

Patol75 (approved):

PR #150, opened 2023-10-05
Add 2D cell test to check adequate convergence for 5000+ grains

adigitoleo:
Adds a 2D cell convection test which is a good way to test numerical convergence (the flow has pure and simple shear components of varying magnitude, and pathlines don't flow into/out of the domain so it can be run until arbitrarily large strain).

The main quantitative diagnostics here are:
- mean angle of olivine a-axis compared to the local velocity direction ("CPO angle")
- maximum grain size (normalized, log10)

The shape of the grain size distribution and the sawtooth shape of the CPO angle are qualitative characteristics of this test.

The test should be able to run without HPC in fact (~8 min, might be over the github CI limit on memory), because in its basic form it doesn't need to be an ensemble: there are features of the texture at high strain which are relatively insensitive to the random state (seed), in particular the dip in both CPO angle and maximum grain size at about 3.75 strain (tensorial). However, I've added an ensemble example which should show that this feature is actually insensitive as I just claimed.

The non-ensemble test produces two figures:
- plot of the velocity field with the pathline, color of pathline dots shows strain
- plot of the diagnostics (CPO angle, and grain sizes at every 10 strain increments)

They look like this:


adigitoleo:
Hold off on merge for a second, I think I might want a slightly tighter test of the CPO angle dip, it seems the one as it is might not catch regression errors very well.

pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "scipy >= 1.2",  # <https://github.com/scipy/scipy/pull/9176>
     "pyyaml",  # For YAML headers of plaintext datafiles.
     "meshio",  # For reading VTK and other meshes.
+    "cmcrameri",  # Matplotlib colormaps using the palettes designed by Fabio Crameri.
```
Patol75, line 41: I would suggest [palettable](https://jiffyclub.github.io/palettable/) instead. It includes more maps.

src/pydrex/pathlines.py
```diff
@@ -1,109 +1,122 @@
...
+    max_strain,
+    **kwargs,
 ):
     """Determine the pathline for a particle in a steady state flow.
 
-    The pathline will intersect the given `point` and follow a curve determined by
-    the interpolated velocity gradient.
+    The pathline will terminate at the given `final_location` and follow a curve
+    determined by the velocity gradient. It works for both 2D (rectangular) and 3D
+    (orthopiped) domains, so long as the provided callables expect/return arrays of the
```
Patol75, line 22: Is orthopiped a word?

src/pydrex/pathlines.py
```diff
@@ -1,109 +1,122 @@
...
-            event_time_prev = event_time
-            event_time = time
-            return event_strain - 10
-
+            dε = _utils.strain_increment(
+                time - _time_prev, get_velocity_gradient(point)
+            )
+            if time > _time_prev:  # Timestamps jump around for SciPy to find the root.
+                _strain += dε
+            else:  # Subtract strain increment becuase we are going backwards in time.
```
Patol75, line 61: because

src/pydrex/visualisation.py
```diff
@@ -131,7 +134,104 @@ def polefigures(
...
+    ax.set_ylabel(ref_axes[1])
+
+    x_min, y_min = min_coords
+    x_max, y_max = max_coords
+    ax.set_xlim((x_min, x_max))
+    ax.set_ylim((y_min, y_max))
+
+    x_res, y_res = resolution
+    if x_res == y_res:
+        ax.set_aspect("equal")
```
Patol75, line 183: Why did you make that choice? You could instead have a threshold on their ratio to include further scenarios. Equal aspect representation is helpful. :)

src/pydrex/visualisation.py
```diff
@@ -321,6 +464,11 @@ def figure_unless(ax):
     return fig, ax
 
 
+def figure():
+    """Create new figure with `dpi=300` and automatic constrained layout, grid etc."""
+    return plt.figure(dpi=300)
```
Patol75, line 469: Why not use `plt.rcParams["figure.dpi"] = 300`? Also, what do you mean by automatic constrained layout? I am not sure it is the default.

Patol75 (changes requested):
Great effort! The test looks super relevant and the results are nicely illustrated!

src/pydrex/pathlines.py
adigitoleo: I went a bit crazy with that, but someone else agrees with me: https://www.whatistoday.net/2020/04/cuboid-dilemma.html

Happy to change it to just 'box' or whatever is best.

src/pydrex/visualisation.py
adigitoleo: Yeah I should put the DPI setting up the top in my rcParams stuff, like how I set up [constrained layout by default](https://github.com/seismic-anisotropy/PyDRex/blob/e0aa2c4b627cc26baf72872844d2be95d9a5d4f3/src/pydrex/visualisation.py#L18), at least for anything that imports `pydrex.visualisation`.

pyproject.toml
adigitoleo: Interesting, I didn't see that one. Feels a bit more clunky to use though, no `batlow`, only `Batlow_20` etc. Not sure I need 700 cmaps, I just want the good ones :) Plus I cannot find a way to forward declare cmaps using string names so I'd have to import the `palettable.scientific.sequential` everywhere.

src/pydrex/pathlines.py
Patol75: As you wish, but if you keep it the way it is, add the URL ;)

Patol75 (approved):

PR #152, opened 2023-10-10
test: Add reworked 2D cornerflow example

adigitoleo:
Also fixes a bug with `misorientation_indices` which caused mixed up M-index outputs and apparent non-reproducible behaviour.

I have removed the old assert statements on this test for now as they don't really test much, also this test is too long for CI anyway and not very tractable as an ensemble test either, so it is mainly qualitative, but at least shows if the M-index is working.

### DIFF
--- a/src/pydrex/diagnostics.py
+++ b/src/pydrex/diagnostics.py
@@ -25,6 +25,7 @@ import scipy.linalg as la
 from pydrex import stats as _stats
 from pydrex import tensors as _tensors
 from pydrex import geometry as _geo
+from pydrex import logger as _log
 
 
 def elasticity_components(voigt_matrices):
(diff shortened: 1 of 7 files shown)


adigitoleo:
Using the multiprocessed version of `misorientation_indices` instead of looping over `misorientation_index` calls cuts the CPU time in half, but this example is the most memory intensive and needs about 60GB (I've added a note in the docstring).

src/pydrex/diagnostics.py
```diff
@@ -289,7 +289,7 @@ def misorientation_indices(
     if ncpus is None:
         ncpus = len(os.sched_getaffinity(0)) - 1
     with Pool(processes=ncpus) as pool:
-        for i, out in enumerate(pool.imap_unordered(_run, orientation_stack)):
+        for i, out in enumerate(pool.map(_run, orientation_stack)):
```
Patol75, line 292: Just wondering if the large memory usage has anything to do with `map`. See [here](https://docs.python.org/3.11/library/multiprocessing.html#multiprocessing.pool.Pool.map).

Patol75 (commented):
Looking great, well done!

src/pydrex/diagnostics.py
adigitoleo: Oh yeah I'll try with `imap` as well, that is still ordered I think.

adigitoleo: Switched to `imap`.

Patol75 (approved):

PR #155, opened 2023-10-18
Fixes to test figure output and multi-platform support

adigitoleo:
- Use PDF for all figures
- Remove redundant old plotting functions
- Use a safe wrapper to get a default number of CPUs for Linux, Mac or Windows
- Make the `venv_install.sh` script support `PYTHON_BINARY` for using custom paths to local Python installations (e.g. 3.11 on cosgrove)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
       - main
 
 jobs:
-  install:
+  install-linux:
     name: (Linux) Install package and run tests
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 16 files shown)


adigitoleo:
Could be useful to add runners for CI on `windows-latest` and `macos-latest` as well, might look at that later tonight.

Patol75 (approved):
Changes look sensible to me :)

Patol75 (dismissed):

Patol75 (approved):

adigitoleo:
Failing on Windows because of strangely large memory commits from scipy integrators (switching back to RK45 doesn't help). Tempted to drop Windows support for now. Potentially related to the SciPy BLAS being different on windows (I think?) or maybe an actual SciPy integrator bug on windows. MacOS is passing.

Patol75:
Perhaps drop Windows for now. It could just be running out of memory. Not quite sure why that would not be the case on Ubuntu though.

Patol75 (approved):

PR #156, opened 2023-10-25
feat: Make redraw_legend more flexible; finalise GBM test figure legend

adigitoleo:
Produces the final version of the GBM vs fortran test as per overleaf manuscript.

The legend was overprinting traces at a font size of 18 so I moved it outside the axes.

### DIFF
--- a/src/pydrex/utils.py
+++ b/src/pydrex/utils.py
@@ -59,7 +59,7 @@ def default_ncpus():
                 return int(out.stdout.strip()) - 1
             case "Windows":
                 return int(os.environ["NUMBER_OF_PROCESSORS"]) - 1
-    except AttributeError or CalledProcessError or KeyError:
+    except AttributeError or subprocess.CalledProcessError or KeyError:
         return 1
 
 
--- a/src/pydrex/utils.py
+++ b/src/pydrex/utils.py
@@ -130,7 +130,7 @@ def redraw_legend(ax, fig=None, remove_all=True, **kwargs):
         arguments `bbox_extra_artists=(legend,)` and `bbox_inches="tight"` to `savefig`,
         where `legend` is the object returned by this function. To prevent the legend
         from consuming axes/subplot space, it is further required to add the lines:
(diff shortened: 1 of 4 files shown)


src/pydrex/utils.py
```diff
@@ -112,12 +115,53 @@ def quat_product(q1, q2):
...
+    Optional keyword arguments are passed to `matplotlib.axes.Axes.legend` by default,
+    or `matplotlib.figure.Figure.legend` if `fig` is not None.
+
+    .. warning::
+        Note that if `fig` is not `None`, the legend may be cropped from the saved
+        figure due to a Matplotlib bug. In this case, it is required to add the
+        arguments `bbox_extra_artists=(legend,)` and `bbox_inches="tight"` to `savefig`,
+        where `legend` is the object returned by this function. To prevent the legend
+        from consuming axes/subplot space, it is further required to add the lines:
+        `legend.set_in_leyout(False)`, `fig.canvas.draw()`, `legend.set_layout(True)`
```
Patol75, line 133: `legend.set_in_layout(False)`

tests/test_simple_shear_2d.py
```diff
@@ -475,7 +480,15 @@ def test_dvdx_GBM(self, outdir, seeds_nearX45, ncpus):
                         "Zhang & Karato, 1995\n(1300°C)",
                     ],
                 )
-                fig.savefig(_io.resolve_path(f"{out_basepath}.pdf"))
+                # There is a lot of stuff on this legend, so put it outside the axes.
+                # The 'loc' argument doesn't seem to work, use `bbox_to_anchor` instead.
```
Patol75, line 484: `loc` and `bbox_to_anchor` work together: `loc` sets the location of the legend box that becomes the target of `bbox_to_anchor`.

Patol75 (changes requested):
Looking reasonable to me.

tests/test_simple_shear_2d.py
adigitoleo: I see, I'll remove that line since it works fine with the default `loc`.

Patol75 (approved):

PR #162, opened 2024-02-05
Add simple f90 performance comparison script (single core)

adigitoleo:
Results:

gfortran 13.2.0
Python 3.11.6 (numba 0.58.1)

(this is gfortran without optimisation)

**n_grains** | **PyDRex** | **DRex**
--- | --- | ---
15^3 | 42s | 42s
17^3 | 58s | 62s
19^3 | 81s | 86s

Oof, with `gfortran -O3` we lose by x10 at least. I'll try to get another quick heatmap to see if something obvious can be done. The new NumPy API where everything now has an `out=foo` array arg might help chaining things together.

### DIFF
--- a/src/pydrex/utils.py
+++ b/src/pydrex/utils.py
@@ -16,7 +16,7 @@ import numpy as np
 def strain_increment(dt, velocity_gradient):
     """Calculate strain increment for a given time increment and velocity gradient.
 
-    Returns “tensorial” strain increment ε, which is equal to 2 × γ where γ is the
+    Returns “tensorial” strain increment ε, which is equal to γ/2 where γ is the
     “(engineering) shear strain” increment.
 
     """


tools/perf_compare.sh
```diff
@@ -0,0 +1,147 @@
...
+
+
+Cij_and_friends = _diagnostics.elasticity_components(
+    _minerals.voigt_averages([mineral], params)
+)
+angles = [
+    _diagnostics.smallest_angle(v, shear_direction)
+    for v in Cij_and_friends["hexagonal_axis"]
+]
+precent_anisotropy = Cij_and_friends["percent_anisotropy"]
```
Patol75, line 137: `percent`

Patol75 (approved):
Good job on that one. It will be a useful tool, and it could even be part of the CI.

PR #163, opened 2024-02-21
MAINT: Update pre-commit and format files

Patol75:
I updated all repo versions and ran `pre-commit` against all files. Should be good to go, but let's see if something terrible happened.


Patol75:
I have the same, and I have not looked into why. Let's merge nevertheless.

PR #169, opened 2024-02-29
fix: Fix IndexError resulting from cumsum(fractions)[-1] != 1

adigitoleo:
This fixes an occasional IndexError that occurs during resampling, if np.cumsum(fractions_ascending) < rng.random(), which results in np.searchsorted() returning an index outside the size of the new arrays.


adigitoleo:
From the `np.cumsum` docs:

> cumsum(a)[-1] may not be equal to sum(a) for floating-point values since sum may use a pairwise summation routine, reducing the roundoff-error. See sum for more information.

Patol75 (approved):

PR #174, opened 2024-03-21
Fluidity advection 2d

adigitoleo:
Add fluidity advection example and simple builtin mesh (gmsh) generation.


adigitoleo:
Failures caused by gmsh not being able to find its dependency:

```
OSError: libGLU.so.1: cannot open shared object file: No such file or directory
```

adigitoleo:
Tracked it down: docs failure is still caused by gmsh import, the workaround from https://github.com/seismic-anisotropy/PyDRex/pull/174/commits/2d30d913c1a90361df7806a12c1248a3d0b9751e doesn't fix it. Since the gmsh package is so much trouble I'll just move it to an optional dep and make the mesh generation functionality unavailable by default.

adigitoleo:
https://github.com/mitmproxy/pdoc/issues/673

adigitoleo:
Manual exclusion of the `mesh` submodule from the docs works for now.

adigitoleo:
Fluidity now seems to produce the correct results. More on this later today. Might want to squash some of this stuff before merge.

Patol75 (dismissed):

Patol75 (approved):

PR #187, opened 2024-04-04
feat: Allow using Ray as a drop-in distributed multiprocessing Pool

adigitoleo:
- Ray has a `multiprocessing.Pool` API that is very similar to the already supported Python stdlib `multiprocessing` stuff
- Refraining from a nested `imap` inside the M-index calculation routines cuts down the memory even more, at the cost of extra CPU time (which can be gained back by running on more nodes now)
- At the moment only the 3d simple shear test is properly hooked up to Ray with the `ray.put` and `ray.get` calls, but doing the same for other `slow` tests would look the same


adigitoleo:
Example:
```sh
#!/bin/bash
#PBS -P xd2
#PBS -q normalbw
#PBS -l walltime=05:00:00
#PBS -l ncpus=168
#PBS -l mem=1536GB
#PBS -l jobfs=800GB
#PBS -l storage=scratch/xd2+gdata/xd2
#PBS -l wd
#PBS -o simple3d_unsteady_prescr_dc1000000.log
#PBS -e simple3d_unsteady_prescr_dc1000000.err
#PBS -N simple3d_unsteady_prescr_dc1000000

# Install pydrex into this python environment.
module purge
module load python3/3.11.7 python3-as-python
ulimit -c unlimited

# Copy the tools/pbs_start_ray*.sh into $PWD.
source pbs_start_ray_cluster.sh

# Run this script from outside the PyDRex tree, which should be in `PyDRex/*`.
ray job submit --working-dir PyDRex/ -- \
    python3 -m pytest tests/test_simple_shear_3d.py \
    --outdir=_out_1 -v --runslow --ncpus=$PBS_NCPUS -k="direction_change[1]"
```

src/pydrex/diagnostics.py
```diff
@@ -279,11 +284,16 @@ def misorientation_indices(
...
-    instance of `multiprocessing.Pool` can be provided.
+    Uses either Ray or the Python multiprocessing library to calculate texture indices
+    for multiple snapshots simultaneously. The arguments `ncpus` and `pool` are only
+    relevant the latter option: if `ncpus` is `None` the number of CPU cores to use is
+    chosen automatically based on the maximum number available to the Python
+    interpreter, otherwise the specified number of cores is requested. Alternatively, an
+    existing instance of `multiprocessing.Pool` can be provided.
+
+    If Ray is installed, it will be automatically preferred. In this case, the number of
+    processors (actually Ray “workers”) should be set upon initialisation of the Ray
```
Patol75, line 295: I do not think that `workers` is specific to Ray. Are you trying to oppose multiprocessing `processes` to Ray `Actors`?

tests/conftest.py
```diff
@@ -152,6 +155,20 @@ def named_tempfile_kwargs(request):
...
+    if HAS_RAY:
+        # NOTE: Expects a running Ray cluster with a number of CPUS matching --ncpus.
+        if not ray.is_initialized():
+            ray.init(address="auto")
+            _log.info("using Ray cluster with %s", ray.cluster_resources())
+        yield None
+        if ray.is_initialized():
+            ray.shutdown()
+    else:
+        yield None
```
Patol75, line 169: Why not just `yield`?

Patol75 (dismissed):
Great effort, looking very good!

src/pydrex/diagnostics.py
adigitoleo: changed wording

tests/conftest.py
adigitoleo: probably copy-pasted from somewhere :)

Patol75 (approved):

PR #191, opened 2024-04-24
MAINT: Update doc actions versions

Patol75:


Patol75:
Looks like Ubuntu is struggling:
```
 Reading package lists...
E: Failed to fetch https://packages.microsoft.com/ubuntu/22.04/prod/dists/jammy/InRelease  Clearsigned file isn't valid, got 'NOSPLIT' (does the network require authentication?)
E: The repository 'https://packages.microsoft.com/ubuntu/22.04/prod jammy InRelease' is no longer signed.
Error: Process completed with exit code 100.
```
Probably related to the release of 24.04. Let's give it some time and trigger again.

adigitoleo (approved):
Thanks

PR #192, opened 2024-04-25
BUG: Investigate doctest Actions failure

Patol75:


Patol75:
So what is happening here is that these [doctsrings examples](https://github.com/seismic-anisotropy/PyDRex/blob/59a5c10afe188ee403629b0795fc91e1636dde2d/src/pydrex/diagnostics.py#L402-L405) return `45.0` on Ubuntu and Windows but `45.0...01` on Mac. This is very likely due to some differences in floating-point-number arithmetic between different platforms. Perhaps the fact that the function goes through Numba's JIT has something to do with it, but `fastmath` does not seem to be responsible.

adigitoleo:
Just using `int()` in this case. If it comes up again when testing against actual floats, I guess we can try `round()`.

Also, maybe we adopt this convention now for commit messages: https://www.conventionalcommits.org/en/v1.0.0/ ? Finally found the link again and it seems like that's what I saw some time ago and was trying to replicate.

adigitoleo:
My `~/.git-commit-template`, maybe I'll add it to the docs:

```

# <type>: <single line subject> (use imperative verbs, start with capital letter,
#                                omit trailing period)
# <multiline body>
# (can use '-' for bullet points, and '[#N]' to reference issue number N)

# Types of commits:
#   fix         (bug fix, append ! if potentially a breaking change)
#   feat        (new feature)
#   remove!     (feature removal)
#   refactor    (internal abstraction change, append ! if potentially a breaking change)
#   style       (formatting/cosmetic change)
#   docs        (documentation change)
#   build       (build system change, dependency change)
#   perf        (performance improvement)
#   test        (test change, fix for broken test)
#   dev         (CI configuration or devtool script change)
#   release     (release commit, should always be tagged)

# Enable this template using:
#   git config --global commit.template <path to template file>
```

adigitoleo:
OK now pytest -v shows one-by-one test progress and logs, and -vv also shows verbose doctest output.

adigitoleo (approved):
:window: :apple:

adigitoleo (approved):

PR #193, opened 2024-05-29
Documentation fixes/improvements

adigitoleo:
Nothing new here, just fixing some documentation. README now suggests installing the PyPi release, elasticity tensor decomposition properly references symmetry of the medium instead of "minerals" (which reads as if it were single-crystal symmetry) and the venv install script sets up a proper dev environment with some ignore rules. First of many PRs in the next while, but I'm cutting them up to be manageable.

### DIFF
--- a/src/pydrex/diagnostics.py
+++ b/src/pydrex/diagnostics.py
@@ -40,7 +40,7 @@ def elasticity_components(voigt_matrices):
 
     Args:
     - `voigt_matrices` (array) — the Nx6x6 Voigt matrix representations of the averaged
-      elasticity tensors for a series of polycrystal textures
+      elasticity tensors for a series of N polycrystal textures
 
     Returns a dictionary with the following data series:
     - `'bulk_modulus'` — the computed bulk modulus for each Voigt matrix C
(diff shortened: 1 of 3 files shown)


src/pydrex/diagnostics.py
```diff
@@ -49,15 +49,15 @@ def elasticity_components(voigt_matrices):
...
+      tetragonally symmetric bulk medium
     - `'percent_orthorhombic'` — for each C, the percentage of ||C|| attributed to
-      orthorhombically symmetric minerals
+      orthorhombically symmetric bulk medium
     - `'percent_monoclinic'` — for each C, the percentage of ||C|| attributed to
-      monoclinically symmetric minerals
+      monoclinically symmetric bulk medium
     - `'percent_triclinic'` — for each C, the percentage of ||C|| attributed to
-      triclinically "symmetric" minerals (no mirror planes)
+      triclinically "symmetric" bulk medium (no mirror planes)
```
Patol75, line 60: You added 'a' before 'hexagonally' but not before other symmetry systems.

Patol75 (changes requested):
Looking good! Just a small consistency issue in the docstring I believe.

Patol75 (approved):

PR #195, opened 2024-06-05
Add diffusion and yielding

adigitoleo:
Some more rebased commits. Adds texture evolution (or lack thereof) for diffusion creep (with tests) and yielding. There are a few commented alternative ways to do diffusion creep, I'd like to leave them in as comments for now in case I come back to it.

Other things:
- simpler way to generate text file listing final particle locations
- peridotite solidus from Herzberg 2000 (with easy way to add other fits)
- get second order tensor invariants (e.g. for strain tensor)
- remove duplicated `PERMUTATION_SYMBOL` in `tensors` (it's in `core`)
- some test/visualisation improvements, requiring `dill` to allow "pickling" lexical closures
- the diffusion creep test (in `test_vortex2d`) relies on the above
- optional mypy depencency (and lsp plugin) for dev environment

### DIFF
--- a/src/pydrex/core.py
+++ b/src/pydrex/core.py
@@ -54,7 +54,7 @@ class DeformationRegime(IntEnum):
     The mechanism of deformation that dominates accommodation of plastic deformation
     depends in general both on material properties such as grain size and mineral phase
     content as well as on thermodynamic properties such as temperature, pressure and
-    water content (via its fugacity).
+    water fugacity.
 
     The activity of diffusive mechanisms depends more strongly on grain size, whereas
     that of dislocation mechanisms depends more strongly on temperature. High
@@ -69,6 +69,10 @@ class DeformationRegime(IntEnum):
     e.g. [Gouriet et al. 2019](http://dx.doi.org/10.1016/j.epsl.2018.10.049),
     [Garel et al. 2020](http://dx.doi.org/10.1016/j.epsl.2020.116243) and
     [Demouchy et al. 2023](http://dx.doi.org/10.2138/gselements.19.3.151).
+
+    .. note:: Although a draft texture evolution behaviour is implemented in the
+        `frictional_yielding` regime, it is experimental and not yet configurable via
+        the parameter interface.
 
     """
 
--- a/src/pydrex/core.py
+++ b/src/pydrex/core.py
@@ -242,6 +242,7 @@ def derivatives(
     fractions,
     strain_rate,
     velocity_gradient,
+    deformation_gradient_spin,
     stress_exponent,
     deformation_exponent,
     nucleation_efficiency,
@@ -259,6 +260,8 @@ def derivatives(
     - `fractions` (array) — volume fractions of the grains relative to aggregate volume
     - `strain_rate` (array) — 3x3 dimensionless macroscopic strain-rate tensor
     - `velocity_gradient` (array) — 3x3 dimensionless macroscopic velocity gradient
+    - `deformation_gradient_spin` (array) — 3x3 spin tensor defining the rate of
+                                            rotation of the finite strain ellipse
     - `stress_exponent` (float) — `p` in `dislocation_density ∝ shear_stress^p`
     - `deformation_exponent` (float) — `n` in `shear_stress ∝ |deformation_rate|^(1/n)`
     - `nucleation_efficiency` (float) — parameter controlling grain nucleation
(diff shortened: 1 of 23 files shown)


adigitoleo:
Also added `@utils.defined_if(cond)` decorator because `itertools.batched` is 3.12 only. The condition can be anything, though, so this could be used to make some of the conditional `gmsh` and `ray` stuff cleaner but I don't want to mess with it at the moment.

src/pydrex/utils.py
```diff
@@ -36,6 +39,59 @@ def import_proc_pool():
...
+        self._f = dill.dumps(f, protocol=5, byref=True)
+
+    def __call__(self, *args, **kwargs):
+        return dill.loads(self._f)(*args, **kwargs)
+
+
+def serializable(f):
+    """Make decorated function serializable.
+
+    .. warning:: The decorated function cannot be a method, and it will loose it's
```
Patol75, line 69: `its`

src/pydrex/visualisation.py
```diff
@@ -234,21 +241,22 @@ def pathline_box2d(
...
-            ]
-        )
-        cpo = ax.quiver(
-            P[:, 0],
-            P[:, 1],
-            C[:, 0],
-            C[:, 1],
+    dummy_dim = ({0, 1, 2} - set(_geo.to_indices2d(*ref_axes))).pop()
+    xi_2D = np.asarray([_utils.remove_dim(p, dummy_dim) for p in positions])
+    qcoll: plt.Quiver | plt.PathCollection
```
Patol75, line 246: Is that a type annotation in the middle of nowhere?

tests/test_vortex_2d.py
```diff
@@ -1,25 +1,132 @@
...
 from pydrex import utils as _utils
 from pydrex import velocity as _velocity
 from pydrex import visualisation as _vis
+from pydrex import geometry as _geo
 
 Pool, HAS_RAY = _utils.import_proc_pool()
 
-# Subdirectory of `outdir` used to store outputs from these tests.
 SUBDIR = "2d_vortex"
+"""Subdirectory of `outdir` used to store outputs from these tests."""
```
Patol75, line 26: Triple quotes?

Patol75 (changes requested):
Great round of changes. As always, a few comments here and there.

src/pydrex/core.py
```diff
@@ -49,12 +49,45 @@ class MineralPhase(IntEnum):
 
 @unique
 class DeformationRegime(IntEnum):
-    """Deformation mechanism regimes."""
+    r"""Ordinals to track distinct regimes of dominant deformation mechanisms.
+
+    The mechanism of deformation that dominates accommodation of plastic deformation
+    depends in general both on material properties such as grain size and mineral phase
+    content as well as on thermodynamic properties such as temperature, pressure and
+    water content (via its fugacity).
```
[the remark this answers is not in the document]
adigitoleo: I should reword it, I guess if anything the "water content" (ppm? number of substituted H and O? are these equivalent?) is a proxy for the fugacity and not the other way around, right?

src/pydrex/core.py
```diff
@@ -234,28 +272,93 @@ def derivatives(
...
-            orientations[grain_index],
-            strain_rate,
-            velocity_gradient,
-            stress_exponent,
-            deformation_exponent,
-            nucleation_efficiency,
+    if regime == DeformationRegime.min_viscosity:
+        # Do absolutely nothing, all derivatives are zero.
+        # TODO: Consider resetting the texture to isotropic, treat it like a phase
+        # change or something?
```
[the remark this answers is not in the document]
adigitoleo: Agreed, the tracker idea seems valuable. I had made a conscious effort to limit the amount of things tracked on `Mineral`s to ensure loose coupling (and avoid excessive memory use), but a history of regimes seems sensible. Instead of an array of values for each timestep, which would end up fairly sparse in most models (matrix_dislocation most of the time) I could do an array of tuples of (regime, duration in timesteps).

src/pydrex/core.py
```diff
@@ -234,28 +272,93 @@ def derivatives(
...
+                stress_exponent,
+                deformation_exponent,
+                nucleation_efficiency,
+            )
+            orientations_diff[grain_index] = 0.3 * orientation_change
+            strain_energies[grain_index] = strain_energy
+        # Volume average mean strain energy.
+        mean_energy = np.sum(fractions * strain_energies)
+        # Strain energy residuals, minus the energy lost to micro-fracturing.
+        strain_residuals = 0.3 * (mean_energy - strain_energies)
```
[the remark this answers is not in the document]
adigitoleo: Source: trust me :) I'll have to find all the papers I looked at and do a proper statistical comparison but for now this value was my qualitative sense for roughly how much energy (30%) would be left. Maybe a note about this regime being experimental, make it a config option... Probably no experiments to really compare CPO to here.

src/pydrex/visualisation.py
adigitoleo: It's valid syntax and covers the dual use of `qcoll` which subsequently collects either the output of `ax.quiver` or `ax.scatter`. Should be transparent to anything but mypy. Type hints now became more generic and less painfull (3.12): https://www.andy-pearce.com/blog/posts/2023/Dec/whats-new-in-python-312-type-hint-improvements/ and I'm starting to see the appeal, although I refuse to import `typing` anywhere so I'm not about to add lots of hints, just here and there to help with code review.

tests/test_vortex_2d.py
adigitoleo: So that the text appears in the generated html

src/pydrex/core.py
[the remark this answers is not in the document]
Patol75: Water content is a general term to designate a measure of water in a system. It could be a concentration, a fugacity (which is a partial pressure if I am not mistaken), an atomic or molecular measure, etc...

Patol75 (approved):

PR #197, opened 2024-06-08
General clean

### LINKED ISSUE #196: Track terse history of deformation regimes on minerals
Regimes are `IntEnum` instances so shouldn't be to memory heavy, but either way we can collapse the history to either a sequence of tuples: `(regime, duration)` or just imply the encoding `[regime, duration, regime, duration, ...]`. Numba may decide for us which of these works. Duration is a number of timesteps, also an int.

Just making an issue so I don't forget. Feel free to merge #195 if the other things are fine, then I'll do this in a quick PR tonight.

adigitoleo:
Some initial prep before the variable visc tests. After this we should be able to restore the [pre-commit ci bot](https://github.com/apps/pre-commit-ci) if we want.

Closes #196 as per the discussion there.

### DIFF
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -124,7 +124,6 @@ def peridotite_solidus(pressure, fit="Hirschmann2000"):
 def voigt_averages(minerals, phase_assemblage, phase_fractions):
     """Calculate elastic tensors as the Voigt averages of a collection of `mineral`s.
 
-    Args:
     - `minerals` — list of `pydrex.minerals.Mineral` instances storing orientations and
       fractional volumes of the grains within each distinct mineral phase
     - `phase_assemblage` — collection of `pydrex.core.MineralPhase`s
@@ -370,10 +369,11 @@ class Mineral:
 
     def update_orientations(
         self,
-        config,
-        deformation_gradient,
+        params: dict,
+        deformation_gradient: np.ndarray,
         get_velocity_gradient,
-        pathline,
+        pathline: tuple,
+        get_regime=None,
         **kwargs,
     ):
         """Update orientations and volume distribution for the `Mineral`.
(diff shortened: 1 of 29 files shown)


src/pydrex/minerals.py
```diff
@@ -381,16 +382,24 @@ def update_orientations(
...
-          returns a 3x3 velocity gradient matrix at time t and position x (3D vector)
-        - `pathline` (tuple) — tuple consisting of:
-            1. the time at which to start the CPO integration (t_start)
-            2. the time at which to stop the CPO integration (t_end)
-            3. callable with signature f(t) that returns the position of the mineral at
-                time t ∈ [t_start, t_end]
+        .. note:: For multi-phase simulations, the return value should only be collected
+            for the last phase, i.e. the last call to `update_orientations`. This value
+            should be passed on to subsequent calls to `update_orientations` as the
+            `deformation_gradient` argument.
```
Patol75, line 388: Just putting a quick comment here that we perhaps need to think of a way to change this.

adigitoleo: Maybe a new `Pathline` object that can store `.x1`, `.x2`[, `.x3`], `.t`, `.strain`, and `.regime`, or anything else that is "macroscopic" history and doesn't fit well as a `Mineral` attribute.

adigitoleo: Could make `get_pathline` return that object instead of the tuple.

Patol75: I will get this merged for now. However, it seems more straightforward to me that a user only needs to call `update_orientations` once. In that sense, this method could be modified to handle multiple phases in one call, or it could be renamed to `update_orientations_single_phase` and the new `update_orientations` would wrap around it and handle the eventual multiple phases. In any case, these approaches would guarantee a single return tuple, which would prevent confusion.

Patol75 (approved):
Good round of updates!

PR #200, opened 2024-06-17
Add multiphase texture update API

adigitoleo:
### feat: Add single-call API for multiphase texture updates

On line 638 of `minerals.py`.

Closes https://github.com/seismic-anisotropy/PyDRex/issues/198. Also rearranges definitions so that globals are above classes are above functions. This allows for mypy type hints of the multiphase function args.

### fix: Update voigt averaging to use new StiffnessTensors API

This fix was brought to you by mypy static type linting.

### DIFF
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -677,6 +677,7 @@ def voigt_averages(
     minerals: list[Mineral],
     phase_assemblage: list[_core.MineralPhase],
     phase_fractions: list[float],
+    elastic_tensors: StiffnessTensors = StiffnessTensors(),
 ):
     """Calculate elastic tensors as the Voigt averages of a collection of `mineral`s.
 
@@ -701,8 +702,6 @@ def voigt_averages(
         raise ValueError(
             "cannot average minerals with variable-length grain volume arrays"
         )
-
-    elastic_tensors = StiffnessTensors()
 
     # TODO: Perform rotation directly on the 6x6 matrices, see Carcione 2007.
     # This trick is implemented in cpo_elastic_tensor.cc in Aspect.
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -651,25 +651,16 @@ def update_all(
 
     """
     for i, mineral in enumerate(minerals):
-        if i == len(minerals) - 1:  # Collect new deformation gradient at last.
-            deformation_gradient = mineral.update_orientations(
-                params=params,
-                deformation_gradient=deformation_gradient,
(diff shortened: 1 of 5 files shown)


adigitoleo:
Failures are from Numpy 2.0 changes. Working on the fix.

src/pydrex/minerals.py
```diff
@@ -712,3 +633,120 @@ def from_file(cls, filename, postfix=None):
...
+            )
+        else:
+            mineral.update_orientations(
+                params=params,
+                deformation_gradient=deformation_gradient,
+                get_velocity_gradient=get_velocity_gradient,
+                pathline=pathline,
+                get_regime=get_regime,
+                **kwargs,
+            )
```
Patol75, line 671: Not even sure you need the `if` here. Could just keep the upper block and update the comment to explain that the deformation gradient is independent of the mineral assemblage. And then rename the returned value to `updated_deformation_gradient`. Up to you.

Patol75 (commented):
Looking good. Just one suggestion. Also, what was the change for `voigt_averages`?

src/pydrex/minerals.py
```diff
@@ -76,109 +100,6 @@ def __iter__(self):
...
-                                mineral.orientations[i][n, ...].transpose(),
-                            )
-                            * mineral.fractions[i][n]
-                            * phase_fractions[phase_assemblage.index(mineral.phase)]
-                        )
-                    case _core.MineralPhase.enstatite:
-                        average_tensors[i] += _tensors.elastic_tensor_to_voigt(
-                            _tensors.rotate(
-                                elastic_tensors["enstatite"],
-                                minerals.orientations[i][n, ...].transpose(),
```
adigitoleo, line 172: Old enstatite code, tries to use a key instead of member, and `minerals` instead of `mineral`.

src/pydrex/minerals.py
```diff
@@ -712,3 +633,120 @@ def from_file(cls, filename, postfix=None):
...
+                                mineral.orientations[i][n, ...].transpose(),
+                            )
+                            * mineral.fractions[i][n]
+                            * phase_fractions[phase_assemblage.index(mineral.phase)]
+                        )
+                    case _core.MineralPhase.enstatite:
+                        average_tensors[i] += _tensors.elastic_tensor_to_voigt(
+                            _tensors.rotate(
+                                elastic_tensors.enstatite,
+                                mineral.orientations[i][n, ...].transpose(),
```
adigitoleo, line 726: New enstatite code.

Patol75 (approved):

PR #201, opened 2024-07-24
Add pydrex.viscosity module

adigitoleo:
First part of adding variable viscosity support. The main things I add here are:
- `pydrex.viscosity` module with functions to get effective viscosity contributions from different mechanisms
- a bunch of references and details to docstrings of the default parameter choices
- a draft of the Fluidity 2D ridge variable viscosity example, but without the actual variable viscosity implemented yet (will add that in a part 2, this flml is just from an older branch)
- support for custom constraints in the gmsh rectangular mesh generator (so I can have high resolution at the top centre, for 2D ridges)

Unfortunately the ridge example doesn't work with the mesh generator for some reason, so I have to use a `.geo` file instead. Might be able to fix this at some point but will focus on getting the full variable viscosity tests validated first.


src/pydrex/core.py
```diff
@@ -140,30 +157,66 @@ class DefaultParams:
...
     disl_prefactors: tuple = (1e-16, 1e-17)
-    """Prefactors for dislocation creep exponential & power laws (s⁻¹)."""
-    diff_prefactor: float = 1e-10
-    """Prefactor for diffusion creep power law (s⁻¹)."""
+    """Prefactors for dislocation creep exponential & power laws (s⁻¹).
+
+    - B = 1.7e-16 s⁻¹ for the exponential law suggested by [Demouchy et al. 2023](http://dx.doi.org/10.2138/gselements.19.3.151)
+    - pre-exponential factor of 4.4e-17 Pa⁻ⁿs⁻¹ used for the exponential law by [Garel et al. 2020](http://dx.doi.org/10.1016/j.epsl.2020.116243)ᵃ
+
+    ᵃCheck units, why is the Pa⁻ⁿ in there...
```
Patol75, line 207: That is normal, it is coming from the general relationship between stress and strain.

src/pydrex/viscosity.py
```diff
@@ -0,0 +1,249 @@
...
+from pydrex.core import DefaultParams
+
+
+def harmonic_mean(viscosities: tuple[float, ...], bounds: tuple[float, float]):
+    """Calculate harmonic mean of `viscosities` and clip result between `bounds`."""
+    # Defensive .clip for each component as well.
+    return np.clip(
+        len(viscosities) / np.sum([1 / np.clip(μ, *bounds) for μ in viscosities]),
+        *bounds,
+    )
```
Patol75, line 32: `scipy.stats.hmean`?

Patol75 (commented):
Looking great so far; another significant contribution!

Patol75 (approved):

PR #205, opened 2024-08-01
Ensure pre-commit hooks pass

Patol75:
These changes should fix the current syntax check failure from CI following pre-commit integration.

### DIFF
--- a/src/pydrex/io.py
+++ b/src/pydrex/io.py
@@ -425,7 +425,7 @@ def save_scsv(file, schema, data, **kwargs):
                             f"invalid data for column '{names[i]}'."
                             + f" Cannot parse {d} as type '{t.__qualname__}'."
                         ) from None
-                    if t == bool:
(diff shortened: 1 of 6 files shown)


src/pydrex/io.py
```diff
@@ -747,7 +747,7 @@ def _parse_scsv_cell(func, data, missingstr=None, fillval=None):
         if fillval == "NaN":
             return func(np.nan)
         return func(fillval)
-    elif func == bool:
+    elif isinstance(func, bool):
```
adigitoleo, line 750: In this case, I wanted to literally check that `func` is the `bool()` builtin function. The `isinstance` check is changing this to check the type of `func`. I think a way to fix it and keep pre-commit happy is to check `func.__qualname__`.

adigitoleo:
Thanks, I think I found the reason for that 1 test failure. I can have a look after lunch unless you beat me to it.

Issue #199, opened 2024-06-17
Document tagged versions as well as latest `main`


adigitoleo:
Pdoc needs some extra steps for documenting multiple versions. Luckily, it seems someone has already come up with a way to do it: https://github.com/mitmproxy/pdoc/issues/470#issuecomment-1562473123. We can use that to either document the most recent tag + main HEAD or all tags + main HEAD.

adigitoleo:
Documentation should include the version number next time it builds on top of a tag (after #207). Need to test this with a `v0.1.0-rc1` tag soon (not necessary to have a new PyPI release until `v0.1.0` proper, but could be nice).

- [ ] Ensure that documentation website includes version number
- [ ] Modify `docs.yml` to build for all tags, not just latest
- [ ] Add version selector to the website (javascript, see linked comment above)

I don't think we need to worry about also building docs for `HEAD` on `main`, building them offline is very simple (c.f. Sphinx, Doxygen, ...)

adigitoleo:
We don't want to store built docs for earlier versions in the repository; at the same time, doing `git checkout` on an earlier version and trying to build the docs is not likely to just work, because (transient) dependencies may have been updated since then (and I don't want to pin fixed versions of all dependencies for each PyDRex version). For example, trying to build the docs for v0.0.1 again now fails because we had NumPy 2.0 come out in the meantime.

Part of the problem is that building the docs with `pdoc` fundamentally requires importing the package (dynamic analysis), but I think that is the same for most Python documentation builders. The options here are: (1) forget about building docs for old versions, since we probably won't be maintaining them with backported patches for such a project or (2) store the built docs for older versions in some persistent online storage somewhere and re-download them into subdirectories of `html/` when deploying the docs.

PR #207, opened 2024-08-01
build: Only deploy docs for latest tag

adigitoleo:
This brings us one step towards having docs for every tagged version. For now (after the next tag push) there should only be one website for the latest tagged version. Earlier versions can be built offline. Progress to #199.

In any case, it is better than the current situation: html documentation is being deployed for every single push on `main`, and the docs website did not indicate a version number anywhere. With these GH workflow adjustments, the docs should still be _built_ for each PR/push to main, allowing to catch errors in the documentation setup as they arise, but the html pages will not deployed to GitHub Pages if there is no tag. The version (obtained from `git describe`) is also indicated clearly on the website landing page and in the API docs sidebar footer.

Related things that I rolled in here:
- new make target to build docs (simply `make html` in the action spec)
- favicon and proper branding (multiple resolutions of logo)
- removed a duplicated sentence from the README
- `docs/template/README.md` to explain how to maintain the template
- fixed `rm -r` -> `rm -rf` in `make clean` recipe


adigitoleo:
> docs should still be built for each PR/push to main

Obviously not happening. Either need to split `docs.yml` into `docs-build.yml` and `docs-deploy.yml`, or somehow get the conditionals to work properly in `docs.yml`.

adigitoleo:
I just realised that based on [this log](https://github.com/seismic-anisotropy/PyDRex/actions/runs/10206968651/job/28240878575?pr=207) maybe I don't need to duplicate the common job steps after all (b2f7d77). I guess the if-check in L63 of docs.yml only evaluates after the "Build" task has completed, so even if we aren't on top of a tag it should still check that `make html` works before aborting. I can remove that commit after review.

README.md
```diff
@@ -45,34 +45,27 @@ declared in the `pyproject.toml` file.
...
-The documentation can be built offline using [pdoc](https://github.com/mitmproxy/pdoc),
-with the command:
+The documentation can be built offline from a git clone or source distribution.
+Install documentation builder dependencies with `pip install '.[doc]'`.
 
-    pdoc -t docs/template -o html --math pydrex tests
-
-which will output the html documentation into a folder called `html`.
+Run the make command `make html` which will output the documentation,
+including the generated API reference, into a folder called `html`.
```
Patol75, line 63: Run `make html` from the terminal to generate PyDRex's documentation (available in the `html` directory), including the API reference.

.github/workflows/docs.yml
```diff
@@ -1,15 +1,39 @@
...
   push:
     branches:
       - main
+    tags:
+      - v*
   pull_request:
     branches:
       - main
-  workflow_dispatch:
+  workflow_dispatch:  # Allows to manually trigger the job from the web UI.
```
Patol75, line 12: You should be able to remove the colon after `workflow_dispatch`.

Patol75 (changes requested):
Looking good; some minor comments.

.github/workflows/docs.yml
adigitoleo: pre-commit fails though, seems like invalid yaml without it:

```
check yaml...............................................................Failed
- hook id: check-yaml
- exit code: 1

while scanning a simple key
  in ".github/workflows/docs.yml", line 12, column 3
could not find expected ':'
  in ".github/workflows/docs.yml", line 14, column 1
```

Patol75 (approved):

PR #213, opened 2024-08-12
Fix logging/pytest setup and make mock objects subclasses of `DefaultParams`

### LINKED ISSUE #211: Use `tmp_path` fixture for more robust temporary files in tests
I just found that pytest provides [`tmp_path`](https://docs.pytest.org/en/stable/how-to/tmp_path.html#tmp-path) which is probably a better way to handle tempfiles in tests compared to the current `tempfile.NamedTemporaryFile` approach, especially because the behaviour of `tempfile` classes is platform-dependent (mainly causing issues on Windows).

adigitoleo:
- Some changes to logging, sending some symbols to `pydrex.io` instead
- Changes to `pydrex.logger` documentation, recommending use of `logging.getLogger("pydrex")` for users (which will not mess with their numpy printoptions)
- more robust pytest live logging plugin and setup, with early errors if the couple of problematic pytest command line options are discovered (and collapsing `-v` and `log_cli=true` into the same behaviour)
- type validation during construction of some dataclasses
- `live_docs` make target, allows browsing up to date documentation while editing
- closes #211 

No `QueueHandler` for multiprocessing just yet.

### DIFF
--- a/src/pydrex/logger.py
+++ b/src/pydrex/logger.py
@@ -57,7 +57,8 @@ In these examples, (open) temporary files and streams are used for demonstration
 
 >>> import tempfile
 >>> import io
->>> tmp = tempfile.NamedTemporaryFile()
+>>> kwargs = { "delete_on_close": False } if sys.platform == "win32" else {}
+>>> tmp = tempfile.NamedTemporaryFile(**kwargs)
 >>> pydrex_logger.debug("debug message")
 >>> with pydrex.io.logfile_enable(io.TextIOWrapper(tmp.file)):  # doctest: +ELLIPSIS
 ...     pydrex_logger.debug("debug message in %s", tmp.file.name)
--- a/src/pydrex/logger.py
+++ b/src/pydrex/logger.py
@@ -57,7 +57,7 @@ In this example, (open) temporary files and streams are used for demonstration.
 
 >>> import tempfile
 >>> import io
->>> kwargs = { "delete_on_close": False } if sys.platform == "win32" else {}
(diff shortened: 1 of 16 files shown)


tests/test_scsv.py
```diff
@@ -98,7 +87,7 @@ def test_read_specfile():
 
 
 @pytest.mark.skipif(_utils.in_ci("win32"), reason="Items are not equal")
```
Patol75, line 89: Should we try to see if this test now passes on Windows?

adigitoleo: I'll check locally on the laptop and remove that skip if it works.

adigitoleo: It still fails for an unrelated reason. There is an extra `\n` element in one but not the other file when doing `readlines()` on Windows for some reason. Will swap to `read().splitlines()` and filtering out empty list elements `''`.

src/pydrex/logger.py
```diff
@@ -1,59 +1,80 @@
...
-    value = 42
-    _log.critical("critical error with value: %s", value)
-    _log.error("runtime error with value: %s", value)
-    _log.warning("warning with value: %s", value)
-    _log.info("information message with value: %s", value)
-    _log.debug("verbose debugging message with value: %s", value)
-    ... # Construct Minerals, update orientations, etc.
-
-```
+.. note:: Always use the old printf style formatting for log messages, not fstrings,
```
Patol75, line 3: Really???

Patol75 (approved):
Great effort!

PR #216, opened 2024-09-16
Improve default LSODA tolerances

adigitoleo:
Loosening the `atol` tolerance for the main LSODA solver improves performance significantly without failing any of our tests. Because we already modify `atol` based on the initial condition, this change mostly just prevents LSODA from trying to meet unreasonably tight error tolerances for those elements of the solution where the corresponding initial condition is zero. With this improvement we go from being sometimes more than 20x slower than the Fortran (for simple shear, excluding Numba compilation time) to being about 7 or 8 times slower.

### DIFF
--- a/src/pydrex/minerals.py
+++ b/src/pydrex/minerals.py
@@ -520,7 +520,7 @@ class Mineral:
             time_start,
             y_start,
             time_end,
-            atol=kwargs.pop("atol", np.abs(y_start * 1e-6) + 1e-12),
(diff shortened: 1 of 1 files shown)


adigitoleo:
Also, changing these components of `atol` to be below 1e-4 doesn't improve things any further, at least from my rough testing. Similarly changing `rtol` doesn't do very much.

Patol75 (approved):

PR #217, opened 2024-09-16
Avoid zero-division under uniaxial compression

adigitoleo:
This fixes a zero-division for some orientations under uniaxial compression, by catching the problematic values for slip invariants and returning null-values for orientation change and grain growth. There are some conditions where slip is simply not possible at all, and it is worth adding this check even though most Real World Problems are unlikely to hit this issue (strain rates would normally have pure shear, simple shear and compressional components combined).

### DIFF
--- a/src/pydrex/core.py
+++ b/src/pydrex/core.py
@@ -714,6 +714,12 @@ def _get_rotation_and_strain(
     """
     crss = get_crss(phase, fabric)
     slip_invariants = _get_slip_invariants(strain_rate, orientation)
+    # Handle the case where all slip invariants are zero,
+    # thus no slip is possible (occurs for a few specific orientations and strain rates,
+    # for example in the case of uniaxial compression and an identity orientation
+    # i.e. grain orientation aligned to the coordinate system).
+    if np.all(slip_invariants == 0):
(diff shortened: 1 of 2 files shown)


src/pydrex/core.py
```diff
@@ -714,6 +714,12 @@ def _get_rotation_and_strain(
     """
     crss = get_crss(phase, fabric)
     slip_invariants = _get_slip_invariants(strain_rate, orientation)
+    # Handle the case where all slip invariants are zero,
+    # thus no slip is possible (occurs for a few specific orientations and strain rates,
+    # for example in the case of uniaxial compression and an identity orientation
+    # i.e. grain orientation aligned to the coordinate system).
+    if min(slip_invariants) == max(slip_invariants) == 0.0:
```
Patol75, line 721: `not slip_invariants.any()` or `np.all(slip_invariants == 0)`
What do you think?

Patol75 (changes requested):
Good catch! Might just change the condition. Up to you.

src/pydrex/core.py
adigitoleo: Sure, changed to the second one.

Patol75 (approved):

PR #224, opened 2025-09-22
fix: Fix unbound custom_points local in mesh generator

adigitoleo:
Fixes error from unbound `custom_points` used in L100 of `cli.py`. The variable should be initialised to `None` before passing CLI arguments in case custom points are not supplied on the command line.

### DIFF
--- a/src/pydrex/cli.py
+++ b/src/pydrex/cli.py
@@ -55,6 +55,7 @@ class MeshGenerator(CliTool):
             center = [float(s) for s in args.center.split(",")]
             assert len(center) == 2
 
+        custom_points = None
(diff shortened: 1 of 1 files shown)


adigitoleo:
Novel CI failure on MacOS only, re-running job to check if it is flaky or persistent

adigitoleo:
Failure is persistent but unrelated to the bugfix (math error due to underlying system libs?) so suggest merging anyway. I don't have a Mac to play with so let's hope it is solved when we bump Python to 3.13 😄

Patol75 (approved):
Easy fix, thank you.

PR #225, opened 2025-09-25
Maintenance tasks related to system and software upgrades

Patol75:
In this PR, I removed maintenance for Python 3.10 and introduced testing for Python 3.13.
Additionally, I made the CI use the latest GitHub Windows runner, swapped the Arm64 version of macOS for the Intel one, and updated the versions of Actions used. 
Finally, I fixed errors and warnings raised when running the test suite with Python 3.13 and recent versions of Python packages.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
 
   install-macos:
     name: (MacOS) Install package and run tests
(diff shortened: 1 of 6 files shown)


adigitoleo (approved):

adigitoleo:
Thanks for these updates! Branch rules should be updated now.

PR #228, opened 2025-11-18
dev: Fix developer venv script to match new minimum Python version re…

adigitoleo:
…quirement


Patol75 (approved):
Tiny, but necessary! Thank you!

